PR #6, opened 2025-03-14
Reformat, add links, and slightly edit README file

mhucka:
In an effort to bring more uniformity to project files, this reformats the file to follow Google Markdown style, with the exception of using 100 character line widths instead of Google's 80.

Additional changes:

- Add links to various things such as Stim docs, Wikipedia pages about the search algorithms, and so on.

- Fix the heading for the usage section, which seems to have been accidentally merged into the first paragraph of that section.

- Change the formating of the basic usage example to avoid using `<` and `>` characters, because those characters have meaning in shell syntax and could have led people to get hard-to-understand errors.

- Make some style edits in some bullet lists, such as not capitalizing the text after colon characters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,4 +140,8 @@ For any questions or concerns not addressed here, please email <quantum-oss-main
 
 ## Disclaimer
 
-Tesseract Decoder is not an official Google product. Copyright 2025 Google LLC.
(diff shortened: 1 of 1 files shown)


LalehB (approved):
LGTM!

PR #16, opened 2025-03-26
Fix #5: remove unused requirements.txt file

### LINKED ISSUE #5: Requirements.txt should be removed if it is not used
The project currently has a `requirements.txt` file at the top level; however, this is file is not being used because the Python interface layer does not exist yet. To avoid misleading visitors into thinking there must be Python code somewhere in here because there's a `requirements.txt` file, I recommend removing the file and adding it back once there's Python code.

mhucka:
Currently there is no Python code in this project, so to avoid user confusion, we may as well remove the `requirements.txt` file until there is some. Resolves issue #5.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,0 @@
-cffi==1.15.1
-mip==1.16rc0
-numpy==1.26.4
-pycparser==2.21
-scipy==1.12.0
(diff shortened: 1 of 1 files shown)


LalehB (approved):
LGTM!

Issue #17, opened 2025-03-27
Add a python interface for Tesseract


LalehB:
it would be nice to add a python interface similar to pymatching (https://github.com/oscarhiggott/PyMatching)

seokhyung-lee:
Hi LalehB, I just created a Python wrapper for the decoder: https://github.com/seokhyung-lee/tesseract-python. Posting it here in case you might be interested!

PR #23, opened 2025-05-17
Add bazel statements needed for pybind11 and export common.h

NoureldinYosri:
This PR:
- modifies `WORKSPACE` to include the statements needed for pybind11
- creates the `pybind_extension` and `py_library` that will wrap tesseract
- exports the classes in `common.h`

this is the first of a series of PRs for https://github.com/quantumlib/tesseract-decoder/issues/17 

---

To check this code you can create a python file `tmp.py`

```py
from src import tesseract_py
s = tesseract_py.common.Symptom([1, 2], 4324)
print(s)
```

and add a build rule
```
load("@rules_python//python:py_binary.bzl", "py_binary")

py_binary(
    name = "tmp",
    srcs = ["tmp.py"],
    deps = [
        ":tesseract_py",
    ],
)
```

```sh
$ bazel run src:tmp
Symptom{D1 D2 }
```


src/BUILD
```diff
@@ -63,6 +65,26 @@ cc_library(
     ],
 )
 
+
+pybind_extension(
+    name = "tesseract_py",
```
noajshu: nit: 
```suggestion
    name = "tesseract_decoder",
```

noajshu (approved):
LGTM, could I suggest the alternate name `tesseract_decoder`?

src/BUILD
NoureldinYosri: done ... I also changed the names of the pybind11 files to match the convention used by stem

NoureldinYosri:
@noajshu I don't have permission to merge

LalehB:
> @noajshu I don't have permission to merge

Oh thanks for mentioning this! I will fix that @NoureldinYosri

PR #24, opened 2025-05-19
BUILD fix for windows

LalehB:


noajshu (approved):
Good idea. When we really want to support windows, that will be a full project bigger than the 1 line in the CI. For now, windows users can use the [WSL](https://learn.microsoft.com/en-us/windows/wsl/install). Thanks!

PR #25, opened 2025-05-26
Replace std::vector<bool> with std::vector<char> for faster computations

draganaurosgrbic:
While profiling tesseract, I noticed bottlenecks in the `to_node` function and `VectorBoolHash` function class when testing **Surface Code Transversal CX Protocols** for larger values of r, d and p (specifically, **r=11,d=11** and **r=13,d=13** and **p=0.002**). A significant percentage of decoding time was spent in code regions operating with `std::vector<bool>` data structures. Using these data structures to store boolean elements can be inefficient (see https://www.geeksforgeeks.org/problem-with-std-vector-bool-in-cpp/ for more detail).

In this PR, I replaced `std::vector<bool>` data structures inside the `TesseractDecoder` class with `std::vector<char>` and retained statements that were assigning true/false literal values to specific elements, as they can be implicitly converted to a char (1 or 0). I also slightly modified the logic for hashing the array of boolean elements in the `VectorBoolHash` function class.

I tested and evaluated the performance impact of this optimization on a **single shot/simulation of a quantum circuit for larger search problems of  Surface Code Transversal CX Protocols**. I performed 10 runs with a fixed seed for generating a shot/simulation and detectors ordering and computed the average decoding time across those runs. 

Note that when performing runs of multiple shots, the optimization will be propagated on each shot and the final result will contain performance improvement aggregated (summed up) across each shot.

Screenshot 2025-05-28 10 49 57 AM
<b>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Performance Impact on Surface Code Transversal CX Protocols (r=13, d=13, p=0.0005, p=0.001, Short Beam)</b>

Screenshot 2025-05-28 10 51 25 AM
<b>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Performance Impact on Surface Code Transversal CX Protocols (r=13, d=13, p=0.0005, p=0.001, Long Beam)</b>

Screenshot 2025-05-28 10 52 53 AM
<b>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Performance Impact on Surface Code Transversal CX Protocols (r=13, d=13, p=0.002, Short Beam)</b>

Screenshot 2025-05-28 10 53 28 AM
<b>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Time Reduction in Key Functions for Surface Code Transversal CX Protocols (r=13, d=13, p=0.002, Short Beam)</b>

### DIFF
--- a/src/tesseract.cc
+++ b/src/tesseract.cc
@@ -85,7 +85,7 @@ void TesseractDecoder::initialize_structures(size_t num_detectors) {
   }
 }
 
-struct VectorBoolHash {
+struct VectorCharHash {
   size_t operator()(const std::vector<char>& v) const {
     size_t seed = v.size(); // Still good practice to incorporate vector size
 
@@ -206,7 +206,7 @@ void TesseractDecoder::decode_to_errors(const std::vector<uint64_t>& detections,
 
   std::priority_queue<QNode, std::vector<QNode>, std::greater<QNode>> pq;
   std::unordered_map<size_t,
-                     std::unordered_set<std::vector<char>, VectorBoolHash>>
+                     std::unordered_set<std::vector<char>, VectorCharHash>>
       discovered_dets;
 
   size_t min_num_dets;
--- a/src/tesseract.cc
+++ b/src/tesseract.cc
@@ -23,9 +23,9 @@ bool Node::operator>(const Node& other) const {
 }
 
 double TesseractDecoder::get_detcost(size_t d,
-                                     const std::vector<bool>& blocked_errs,
+                                     const std::vector<char>& blocked_errs,
                                      const std::vector<size_t>& det_counts,
-                                     const std::vector<bool>& dets) const {
+                                     const std::vector<char>& dets) const {
   double min_cost = INF;
   for (size_t ei : d2e[d]) {
     if (!blocked_errs[ei]) {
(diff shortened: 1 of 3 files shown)


src/tesseract.cc
```diff
@@ -86,14 +86,16 @@ void TesseractDecoder::initialize_structures(size_t num_detectors) {
 }
 
 struct VectorBoolHash {
```
LalehB: since this is not a hash for a `vector<bool>` what do you think about changing the name to VectorCharHash instead of VectorBoolHash?

LalehB (approved):
LGTM! thanks Dragana! 
can you also attach some of your performance plots to this PR?

src/tesseract.cc
draganaurosgrbic: @LalehB Done!

LalehB: @draganaurosgrbic great!

draganaurosgrbic:
@LalehB graphs added!

PR #32, opened 2025-06-11
link tesseract with stim and add tests for the python wrapper of common.h

NoureldinYosri:
some classes in stim are not exposed directly to python (e.g. DemTarget) instead they are exposed through a child class (e.g. ExposedDemTarget) ... this forces us to do the conversion to be able to cross the python/C++ boundary.

---

note that This PR doesn't actually link with stim instead the rule "@stim_py//:stim" rebuilds the python wrapper for stim in a way that allows us to use it.  this is in order to unblock this project until I figure out how to link to the `stim.so` file generated by https://github.com/quantumlib/Stim/blob/f566b83c5da89ab94d7280178e2bd642350180c3/BUILD#L90

TODO: find the correct way to link to the `stim.so` file.

---

part of https://github.com/quantumlib/tesseract-decoder/issues/17

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -65,3 +65,11 @@ http_archive(
     urls = ["https://github.com/bazelbuild/platforms/archive/refs/tags/0.0.6.zip"],
     strip_prefix = "platforms-0.0.6",
 )
+
+http_archive(
+    name = "stim_py",
+    build_file = "//external:stim_py.BUILD",
+    sha256 = "95236006859d6754be99629d4fb44788e742e962ac8c59caad421ca088f7350e",
+    strip_prefix = "stim-1.15.0",
+    urls = ["https://github.com/quantumlib/Stim/releases/download/v1.15.0/stim-1.15.0.tar.gz"],
+)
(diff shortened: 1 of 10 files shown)


WORKSPACE
```diff
@@ -65,3 +65,11 @@ http_archive(
     urls = ["https://github.com/bazelbuild/platforms/archive/refs/tags/0.0.6.zip"],
     strip_prefix = "platforms-0.0.6",
 )
+
+http_archive(
+    name = "stim_py",
```
LalehB, line 70: I have a general question Nour! 
do you think there might be a Stim version mismatch between this line that requires Stim 15 and the one in lines 21–26 of the same WORKSPACE file?

LalehB (approved):
Awesome 👏🏻 
Thank you Nour

WORKSPACE
NoureldinYosri: technically yes, but the difference between these two versions is two days ... for now this is fine since this is a workaround ... I think I should add the bazel rules in external/stim_py to stim itself to avoid having two dependencies on stim

PR #33, opened 2025-06-12
Expose all of the remaining parts of tesseract to python

NoureldinYosri:
with this PR all of tesseract* can be used in python through bazel run/test. I also added string representations to the config classes

in the next PR I will add the ability to do `pip install`

*all of tesseract except the batching method `decode_shots` because it uses `stim::SparseShot` which stim doesn't export to python

---

part of https://github.com/quantumlib/tesseract-decoder/issues/17

### DIFF
--- a/src/common.pybind.h
+++ b/src/common.pybind.h
@@ -1,57 +1,75 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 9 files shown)


src/simplex.cc
```diff
@@ -21,6 +21,17 @@
 
 constexpr size_t T_COORD = 2;
 
+std::string SimplexConfig::str() {
```
LalehB, line 24: nice! thanks for adding this!

LalehB (approved):
LGTM!
Great job Nour 👏🏻

PR #34, opened 2025-06-16
Accelerating 'get_detcost' function

draganaurosgrbic:
Function `get_detcost` is a bottleneck, especially in bivariate-bicycle codes and color codes. This function is frequently called inside the two major loops of the decoding process, when Tesseract computes the cost of a next search state. 

The initial implementation of the function receives two major arguments - vectors `blocked_errs` and `det_counts` that have the same `num_errors` number of elements. The function accesses errors from these two vectors at arbitrary indices, which are indices of errors connected with the detector for which the function value is being evaluated. In addition, the function accesses the `likelihood_cost`  field for each selected error using the `errors` vector, which stores each error as an `common::Error` object, computes the cost for each selected error and returns the minimum cost representing the cost of the detector.

Since memory locations in `blocked_errs` and `det_counts` are accessed at arbitrary, non-sequential indices, this can lead to frequent cache misses. However, the initial implementation overlooks an important detail: within each loop iteration, elements from both `blocked_errs` and `det_counts` are accessed at the same arbitrary index. This means that when one fetches an element from `blocked_errs`, potentially incurring a cache miss, it can also pre-fetch the corresponding element from `det_counts` at no additional cache penalty. This can be achieved by reorganizing these values into a structure that enhances data locality, specifically, creating a structure where `blocked_errs` and `det_counts` are adjacent fields. Furthermore, the design of the structure should be such that its total size aligns well with CPU cache lines (typically 64 bytes on most modern architectures).

I also implemented another optimization strategy called `Early Exit from Loop`. The loop inside the `get_detcost` function computes a cost for each error for the specific detector and returns the minimum cost as the detector cost. If inside the initialization phase we compute the lower bound for the error cost for each error, we can use these lower bound values when computing the detector cost. If we sort errors for each detector by their lower bound cost and inside the `get_detcost` loop we realize that the cost of the current error is at least the minimum cost we computed so far, we can exit the loop (as the lower bound of all the errors following the current error can't have a lower cost than the current we computed). I use a similar technique as I did with `blocked_errs` and `det_counts` to improve the data locality, by storing lower bound values for error costs (called `min_cost`) and the likely cost (called `likelihood_cost`) inside a single structure and fetching their values inside the `get_detcost` function, instead of fetching entire `common::Error` objects.

Below are graphs that evaluate the performance improvement I achieved using these two strategies for accelerating the `get_detcost` function. I present the results for the bivariate-bicycle codes and color codes. I evaluated the improvement in the execution time (seconds) and the improvement in cache misses (percent of cache misses spent in the `get_detcost` function). **For NLR10 bivariate-bicycle codes, I achieved an improvement of more than 50% in the execution time and an improvement of almost 50% for the cache misses. For color codes, I achieved almost 40% improvement in the execution time and more than 70% improvement for the cache misses.**

Note: values in brackets on the graph represent the performance improvement for a specific combination of optimization strategies.

image

image

image

image

image

image

image

image

### DIFF
--- a/src/tesseract.cc
+++ b/src/tesseract.cc
@@ -70,7 +70,6 @@ std::string Node::str()
   ss << "dets=" << self.dets << ", ";
   ss << "cost=" << self.cost << ", ";
   ss << "num_dets=" << self.num_dets << ", ";
-  ss << "blocked_errs=" << self.blocked_errs << ")";
   return ss.str();
 }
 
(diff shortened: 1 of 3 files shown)


LalehB:
@draganaurosgrbic amazing plots and speedup! I was wondering if these cache misses are cache misses of the total program or only the `get_detcost` program because it says "cache misses in the get_detcost function" in the title of plots

LalehB (approved):
LGTM!
Thanks Dragana. very nice speedups 👏🏻

PR #35, opened 2025-06-17
Fixing `-DNDEBUG` handling in HiGHS dependency

noajshu:
Previously we set `-DNDEBUG` define globally in the `.bazelrc`. This is because HiGHS is 1.6x slower if compiled with assertions. However, this also disables assertions in Tesseract and Simplex user code. A better approach is taken here where we compile HiGHS library with a custom BUILD that adds the `-DNDEBUG`


LalehB (approved):
LGTM!
Thanks Noah for fixing this.

PR #44, opened 2025-06-30
make the bfs det order default

LalehB:
It seems that the bfs det order is doing a nice job in practice, so we move it to be the default and one can disable it with `--no-det-order-bfs`


noajshu (approved):
Great idea! yes it is much better.

PR #27, opened 2025-06-03
Removing unnecessary std::vector copy operations

draganaurosgrbic:
### Performance Optimization: Eliminating Unnecessary `std::vector` Copy Operations

This Pull Request resolves a significant performance bottleneck by removing unnecessary `std::vector` copy operations during the decoding process. This fix, which was identified through extensive profiling, provides substantial decoding speedups across various code families, especially in multi-shot simulations.

---

### The Problem: Redundant Vector Copying

When generating new search states, _Tesseract_'s decoding loop was making a copy of a large `std::vector` of blocked errors with the line `next_next_blocked_errs = next_blocked_errs`.

My profiling revealed that this single line was a significant performance bottleneck. For example, in a benchmark on a Color Code, this one line accounted for over **97% of all `std::vector` copy operations**, consuming **20% of the total decoding time** for the entire benchmark (the entire decoding time was 236.66 seconds, and the time spent in this critical line was 46.4 seconds).

The core issue: these expensive copy operations are only necessary when the `--at-most-two-errors-per-detector` heuristic flag is enabled. The current implementation was performing these copies unconditionally, leading to a massive amount of wasted time when the flag was disabled.

Screenshot 2025-07-25 9 41 55 PM

---

### The Solution: Conditionalizing Copy Operations

This PR fixes this bug by making the vector copy conditional. The code now only performs the copy operation when the `--at-most-two-errors-per-detector` flag is active, aligning the implementation with the heuristic's control logic. This simple change eliminates a major source of redundant computation without affecting correctness.

---

### Performance Results

I conducted extensive benchmarks on several code families and different shot counts to quantify the impact of this fix. The results show significant speedups when the heuristic flag is disabled:

* **Transversal CNOT Protocols:** Speedups up to **40.1%**
* **Surface Codes:** Speedups up to **27.0%**
* **Color Codes:** Speedups up to **25.9%**
* **Bivariate-Bicycle Codes:** Speedups up to **19.7%**

(See the included benchmark graphs for a complete overview of the performance improvements.)

These results demonstrate that a single line of unnecessary code can have a profound impact on performance, especially in iterative, computationally-intensive algorithms like _Tesseract_'s decoder.

---

### Key Contributions

* **Identified and Eliminated a Performance Bottleneck:** Pinpointed and removed a critical bug where unnecessary `std::vector` copy operations were consuming a significant portion of decoding time.
[...]

PR #45, opened 2025-07-07
Fixing the performance issue/bug in copying large vectors when --at-most-two-errors-per-detector flag enabled

draganaurosgrbic:
### Fixing a Performance Bottleneck and a Related Bug When Using the `--at-most-two-errors-per-detector` Flag

This PR addresses a critical performance issue and a related bug that existed when the `--at-most-two-errors-per-detector` flag was enabled. The core of the problem was the highly inefficient copying of large `std::vector` data structures, which was necessary for state management (mentioned in #27) but became prohibitively expensive due to previous optimizations. This PR completely removes these costly copy operations and, in doing so, implicitly fixes a bug that was degrading decoder accuracy.

---

### The Performance Issue (and also a Bug)

This PR resolves a performance degradation that was exclusively present when using the `--at-most-two-errors-per-detector` flag. While the code path for these redundant copy operations existed from the beginning, its performance impact was initially masked by other bottlenecks. After significant speedups were achieved in other parts of the code through optimizations in PRs #25, #27, and #34, this flag-specific degradation became a major and prominent bottleneck. The cost of the copy operations escalated as the size of the data structures increased with these PRs. The current work focuses on resolving this last remaining degradation, ensuring that even when this flag is enabled, we achieve a high level of speedup consistent with other scenarios.

---

### The Technical Solution

To solve this, I replaced the expensive copy-and-revert strategy with a more intelligent state management method. Instead of making a copy, I now store a special value (`2`) in the `error_blocked` field of the `DetectorCostTuple` struct. This special value indicates that an error was blocked specifically due to the `--at-most-two-errors-per-detector` flag and should be unblocked (or "reverted") in the next search state. This is made possible because the `error_blocked` field is a `uint32_t`, allowing it to hold values beyond a simple `true/false` (`1` or `0`).

This new approach completely removes the need for the `next_next_detector_cost_tuples` vector. Instead, all changes are now made on a single vector (`next_detector_cost_tuples`), which is much more efficient.

---

### The Bug and its Fix

The old code contained a bug that was degrading the decoder's accuracy. The following loop:

```cpp
for (int d : edets[ei]) {
  next_detectors[d] = !next_detectors[d];
  int fired = next_detectors[d] ? 1 : -1;
  next_num_detectors += fired;
  for (int oei : d2e[d]) {
    next_detector_cost_tuples[oei].detectors_count += fired;
  }

  if (!next_detectors[d] && config.at_most_two_errors_per_detector) {
    for (size_t oei : d2e[d]) {
      next_next_detector_cost_tuples[oei].error_blocked = true;
    }
  }
}
```

had an inconsistency when the `--at-most-two-errors-per-detector` flag was enabled. The number of fired detectors was being updated on `next_detector_cost_tuples`, but the blocking of errors was being performed on `next_next_detector_cost_tuples`. When the `get_detcost` function was called, it would use `next_next_detector_cost_tuples`, which did not have the correct fired detector counts. This inconsistency explains why some benchmarks produced low confidence results and errors (e.g., 3 errors in a Surface Code benchmark).

The new code, which replaces the above loop, is as follows:

```cpp
for (int d : edets[ei]) {
  next_detectors[d] = !next_detectors[d];
  int fired = next_detectors[d] ? 1 : -1;
  next_num_detectors += fired;
  for (int oei : d2e[d]) {
    next_detector_cost_tuples[oei].detectors_count += fired;
  }

  if (!next_detectors[d] && config.at_most_two_errors_per_detector) {
    for (size_t oei : d2e[d]) {
      next_detector_cost_tuples[oei].error_blocked = next_detector_cost_tuples[oei].error_blocked == 1 ? 1 : 2;
    }
  }
}
```

By completely removing the `next_next_detector_cost_tuples` vector and performing all modifications on a single array, this PR implicitly fixes the bug and ensures consistency between the fired detector counts and the blocked error flags.

---

### Benchmarking the Performance and Bug Fix

The results below clearly show a huge amount of time was lost on these expensive copy operations when using the `--at-most-two-errors-per-detector` flag. This was especially apparent in the Surface Code benchmark (r=11, p=0.002, 500 shots). This performance degradation escalated as the size of the data structures increased with previous optimizations.

The bug fix also improved accuracy. For the Surface Code benchmark (r=11, p=0.002, 500 shots) that previously had 3 errors, the new code produced 0 errors.

img1

img2

img3

img4

---

### Analyzing the Impact of the Flag Itself

With the performance and correctness issues fixed, I performed additional experiments to analyze the intrinsic impact of the `--at-most-two-errors-per-detector` flag. The goal was to understand its effect on performance and accuracy now that the underlying implementation is optimized.

The results from these benchmarks show a consistent and strange behavior: the flag provides somewhat better accuracy but lower performance. This is the opposite of a typical heuristic, which is meant to trade some accuracy for a performance gain.

img5

img6

img7

img8

To collect more comprehensive data, I performed additional experiments on various groups of code families. These experiments confirmed the initial findings: the flag provides better accuracy at the cost of performance, with a decoding slowdown ranging from 0.2% to 69%.

img9

img10

img11

img12

img13

img14

img15

img16

---

### Conclusion: This Heuristic is not an Optimization

I performed a final benchmark on a specific case that previously showed a performance benefit when using the flag. The command used for this benchmark was:

`bazel build src:all && time  ./bazel-bin/src/tesseract --pqlimit 200000 --beam 5 --num-det-orders 20  --sample-num-shots 20 --det-order-seed 13267562 --circuit testdata/colorcodes/r=9,d=9,p=0.002,noise=si1000,c=superdense_color_code_X,q=121,gates=cz.stim --sample-seed 717347  --threads 1  --print-stats`

Before the optimizations in PR #34, the execution time without the flag was 75.91 seconds, while using the flag was 74.23 seconds. However, after applying the `get_detcost` optimizations from PR #34, the execution time without the flag was 69.01 seconds, while with the flag it was 72.98 seconds.

This demonstrates that the speedup from the optimizations in PR #34 (a gain of ~6.9 seconds) is far more significant than the speedup this heuristic flag initially provided (a gain of ~1.68 seconds). The flag's intended purpose of improving performance by pruning the search space is now outweighed by the efficiency gains in the core decoding function. The optimization I implemented in PR #34 achieved a better speedup than the one this heuristic flag might initially have achieved, making the flag-specific degradation an issue that needed to be addressed. It is clear that my optimization work now provides a high speedup in all scenarios.

My conclusion is that the current version of the _Tesseract_ algorithm is faster without using this flag. The next logical step may be to remove the flag entirely, but I am leaving that decision to the original implementers of the flag.

---

### Key Contributions

* **Performance Fix:** Removed a performance degradation that was left in the code logic for the `--at-most-two-errors-per-detector` flag. This logic performed redundant copy operations on large vectors instead of a smarter method of reverting changes. This fixes the last remaining performance degradation that was specific to using the heuristic flag.
* **Bug Fix:** Identified and fixed a bug that existed when the `--at-most-two-errors-per-detector` flag was enabled, which was causing accuracy issues and low confidence results.
* **Intelligent Solution:** Implemented a smarter, copy-free state management strategy using a special `2` value in the `DetectorCostTuple` structure, made possible by previous data representation changes.
* **Extensive Benchmarking:** Conducted extensive experiments to evaluate the performance and bug fixes and to analyze the intrinsic behavior of the flag itself.
* **Informed Future Development:** Provided comprehensive data that clearly shows the flag no longer provides a performance benefit and that its behavior is contrary to its purpose as a heuristic. This provides a strong basis for the team to decide on its future.

### DIFF
--- a/src/tesseract.cc
+++ b/src/tesseract.cc
@@ -82,7 +82,6 @@ double TesseractDecoder::get_detcost(
       error_cost = ec.likelihood_cost / dct.detectors_count;
       if (error_cost < min_cost) {
         min_cost = error_cost;
-        // min_error_index = ei;
       }
     }
   }


src/tesseract.cc
```diff
@@ -17,6 +17,7 @@
 #include <algorithm>
 #include <cassert>
 #include <iostream>
+#include <random>
```
LalehB: Which part of the code uses this include?

src/tesseract.cc
```diff
@@ -66,19 +67,25 @@ bool Node::operator>(const Node& other) const {
   return cost > other.cost || (cost == other.cost && num_detectors < other.num_detectors);
 }
 
-double TesseractDecoder::get_detcost(
-    size_t d, const std::vector<DetectorCostTuple>& detector_cost_tuples) const {
+double TesseractDecoder::get_detcost(size_t d,
+                                     const std::vector<DetectorCostTuple>& detector_cost_tuples) {
   double min_cost = INF;
```
LalehB, line 71: Why was the `const` qualifier removed from the function prototype?

src/tesseract.h
```diff
@@ -98,8 +98,10 @@ struct TesseractDecoder {
   size_t num_errors;
   std::vector<ErrorCost> error_costs;
 
+  std::vector<std::vector<int>> d2e_detcost;
+
   void initialize_structures(size_t num_detectors);
-  double get_detcost(size_t d, const std::vector<DetectorCostTuple>& detector_cost_tuples) const;
+  double get_detcost(size_t d, const std::vector<DetectorCostTuple>& detector_cost_tuples);
```
LalehB: same question about `const`?

src/tesseract.cc
```diff
@@ -399,17 +404,17 @@ void TesseractDecoder::decode_to_errors(const std::vector<uint64_t>& detections,
 
         if (!next_detectors[d] && config.at_most_two_errors_per_detector) {
           for (size_t oei : d2e[d]) {
-            next_next_detector_cost_tuples[oei].error_blocked = true;
+            next_detector_cost_tuples[oei].error_blocked =
+                next_detector_cost_tuples[oei].error_blocked == 1 ? 1 : 2;
```
LalehB: It would be helpful to add a comment in the code explaining why 2 is added here.

LalehB (commented):
Thanks, Dragana!
Could you include some benchmarking results for both performance and accuracy? It would be good to quantify the impact of this change.

src/tesseract.cc
draganaurosgrbic: @LalehB This was left here from my work on the new heuristic. I removed it.

src/tesseract.h
draganaurosgrbic: @LalehB I reverted it.

src/tesseract.cc
draganaurosgrbic: @LalehB Done.

src/tesseract.cc
draganaurosgrbic: @LalehB This was left here from my work on the new heuristic. I reverted it.

draganaurosgrbic:
@LalehB see updated description of the PR. It should address all your questions.

LalehB:
> @LalehB see the updated description of the PR. It should answer all of your questions.

Thank you, @draganaurosgrbic, for addressing the comments. I noticed that you're reporting the number of low-confidence events to quantify accuracy — could you also include the error count for completeness?
Additionally, I was wondering how much enabling the `--at-most-two-errors-per-detector` flag impacts accuracy in these benchmarks. Have you observed any noticeable differences with or without this flag?

noajshu:
This is a pretty incredible speedup when the flag is enabled!

Although the flag does effect accuracy, I wonder if we couldn't use a similar technique but possibly allow more than 2. Like `--at-most-4-errors-per-detector` since that's going to have a much smaller impact on accuracy.

I also wonder just how much accuracy is lost for e.g. the large color codes.

noajshu:
> I noticed that you're reporting the number of low-confidence events to quantify accuracy — could you also include the error count for completeness?

+1 could you share the error rates as well please?

draganaurosgrbic:
@LalehB I added the plots that compare the speed and accuracy with and without using the `--at-most-two-errors-per-detector` flag

draganaurosgrbic:
@LalehB see the updated description of the PR, it explains the change/improvement in the accuracy of the decoder. **There was a bug before this PR** (that was occurring only when using the `--at-most-two-errors-per-detector` flag). This PR also fixed this bug.

src/tesseract.cc
```diff
@@ -69,16 +69,22 @@ bool Node::operator>(const Node& other) const {
 double TesseractDecoder::get_detcost(
     size_t d, const std::vector<DetectorCostTuple>& detector_cost_tuples) const {
   double min_cost = INF;
+  double error_cost;
   ErrorCost ec;
   DetectorCostTuple dct;
+  // int min_error_index = -1;
```
LalehB:
```suggestion

```

src/tesseract.cc
```diff
@@ -69,16 +69,22 @@ bool Node::operator>(const Node& other) const {
...
     if (ec.min_cost >= min_cost) break;
+
+    dct = detector_cost_tuples[ei];
     if (!dct.error_blocked) {
-      double error_cost = ec.likelihood_cost / dct.detectors_count;
-      min_cost = std::min(min_cost, error_cost);
+      error_cost = ec.likelihood_cost / dct.detectors_count;
+      if (error_cost < min_cost) {
+        min_cost = error_cost;
+        // min_error_index = ei;
```
LalehB:
```suggestion
```

draganaurosgrbic:
@LalehB Yes, when I performed these experiments/benchmarks, I was getting slower decoding times with this flag, but somewhat higher accuracy (check the graphs for the accuracy/low confidence count)

draganaurosgrbic:
@LalehB Please also note two important contributions of this PR: completely fixed the "bug" of copying large vectors (including when this flag is enabled) and fixed the bug that I didn't know existed before (inconsistent update on `next_detector_cost_tuples` and `next_next_detector_cost_tuples`).

LalehB:
> @LalehB Yes, when I performed these experiments/benchmarks, I was getting slower decoding times with this flag, but somewhat higher accuracy (check the graphs for the accuracy/low confidence count)

@draganaurosgrbic I thought the accuracy improvement was because of the bug fix of "(inconsistent update on next_detector_cost_tuples and next_next_detector_cost_tuples)", however would you please explain to me how the flag `--at-most-two-errors-per-detector` results in worse performance since technically it suppose to do less work? if it makes the performance worse shouldn't we just remove the flag then (since it is introduced to improve performance) and just fix the bug?

draganaurosgrbic:
@LalehB there are two accuracy change/improvement I analyzed. The first one is before and after I fixed the bug (inconsistent update on next_detector_cost_tuples and next_next_detector_cost_tuples). Another analysis is the comparison of accuracy after I fixed this bug: comparison of `--at-most-two-errors-per-detector` after I fixed the bug with the version of not using the flag at all. I have plots/graphs for both of these.

draganaurosgrbic:
@LalehB The first analysis/comparison of accuracy is comparison before and after I fixed the bug. I realized there is a bug because when I completely removed the redundant `std::vector` copy operations I compared the accuracy with the version before I implemented this change. Then I noticed there is a difference in the accuracy. Then I realized there was this bug that we did not notice. I fixed it and explained the difference in the accuracy. After I fixed the bug, I compared the accuracy with and without using the flag. Hope this make sense.

draganaurosgrbic:
@LalehB Thank you so much for your feedback:

- Discovering a bug (especially for someone who worked on the project for a few months) is not easy.
- I did not start this PR with the plan to fix a bug. I noticed a performance issue and decided to fix it.
- When I discovered a performance issue and fixed it, I realized that by fixing that performance issue I also fixed a bug we did not notice before. I realized that by comparing the accuracy before and after fixing performance the issue (and also, implicitly fixing the bug). I documented that, as it is important to understand the change in accuracy and everything that my code changes are affecting.
- After I optimized the algorithm when using this flag, I wanted to understand the impact of this flag on performance and accuracy.
- I did not introduce this flag. This flag existed before. I only discovered a performance issue when using this flag, fixed that performance issue and when fixing that performance issue also fixed the bug that existed and documented everything.
- The only thing missing in this PR might be understanding why this flag actually does not provide better performance, but only accuracy. For this, I need help from people that initially implemented this flag and came up with the idea.
- I tested and benchmarked when using this flag and not using it before I made any changes to Tesseract (before I started my internship). This flag had worse performance even before I started my internship.

draganaurosgrbic:
@LalehB 
"I think here the the impact of bug fix and the heuristic flag --at-most-two-errors-per-detector are mixed up"

No, there are two logical parts of this PR:
1. Fixing the performance issue (which also fixed the bug)
2. After we actually fixed the `--at-most-two-errors-per-detector`, we want to know its impact on performance and accuracy. The results I have shown this flag does not provide better performance, but only better accuracy

draganaurosgrbic:
@LalehB Essential takeaways:

- The code is optimized (and implicitly bug removed)
- Now that we have optimized and code without bugs, I wanted to analyze the performance of this flag after I implemented other optimizations on Tesseract
- My results show this flag has worse performance (at least on the benchmarks I performed)
- Does the rest of the team have experiments they performed when they initially implemented this flag where they noticed better performance? Should I test other benchmarks?

draganaurosgrbic:
@LalehB 

"Is the accuracy improvement coming from the heuristic itself, or from a bug fix?"

This PR contains explanations and graphs for analysis and explanation of accuracy improvement coming both from bug fix and the heuristic itself

draganaurosgrbic:
@LalehB 

"If it's due to the bug fix and not the heuristic, we need to evaluate the heuristic on its own."

Yes, I performed this, after I fixed the flag. I evaluated the heuristic on its own and compared to the code that is not using the flag.

draganaurosgrbic:
@LalehB these graphs show the accuracy improvement coming from the performance fix (and also bug fix, implicitily)

Screenshot 2025-07-15 6 31 31 PM

Screenshot 2025-07-15 6 31 50 PM

draganaurosgrbic:
@LalehB and these graphs from the accuracy improvement coming from the heuristic itself

Screenshot 2025-07-15 6 32 50 PM

Screenshot 2025-07-15 6 33 05 PM

draganaurosgrbic:
@LalehB hope this makes sense now. Always feel free to reach out if I need to provide more clarity. All of these graphs are already included in the PR

draganaurosgrbic:
@LalehB and these graphs show the comparison of performance (decoding speed) without and with using the flag.

Screenshot 2025-07-15 6 37 23 PM

Screenshot 2025-07-15 6 38 09 PM

draganaurosgrbic:
@LalehB I changed the workflow of my explanation in the PR description. Should be more clear now

draganaurosgrbic:
@LalehB also please note these two things:

- I started this PR with the plan to include only the first header/section (**Fixing the performance issue (and also implicitly a bug that existed)**)
- I added the second section (**Analyzing the impact of this flag on performance and accuracy**) because you asked/requested ("Additionally, I was wondering how much enabling the --at-most-two-errors-per-detector flag impacts accuracy in these benchmarks. Have you observed any noticeable differences with or without this flag?")

draganaurosgrbic:
@LalehB @noajshu The PR description is updated, including comprehensive data for analyzing the impact of the flag on performance and accuracy, summary list that includes major contributions, as well as the results for the benchmark we looked at during our last meeting.

LalehB (approved):
LGTM!

PR #46, opened 2025-07-09
Caching and trimming errors from 'get_detcost' function

draganaurosgrbic:
TODO


draganaurosgrbic:
@LalehB I will update the conclusion for those two. I intentionally left those two examples in the results, as they exhibited longer decoding times due to not being able to converge in the selected priority queue limit, but the new heuristic clearly showed that it speeds up the `get_detcost` even more compared to other optimizations I implemented

draganaurosgrbic:
@LalehB description updated

PR #51, opened 2025-07-15
fix naming issue that caused pypi to mismatch wheels

NoureldinYosri:
tested in this testpypi prerelease https://test.pypi.org/project/tesseract-decoder/0.1.2.dev20250715155105/


LalehB (approved):
LGTM!
thank you Nour!

PR #56, opened 2025-07-19
Use rooted URL for the image in SECURITY.md

mhucka:
Turns out the security policy document is accessible on GitHub from at least two paths, and the relative image URL only worked in one and not the other. Using a rooted URL fixes that problem.


LalehB (approved):
LGTM!
Thank you Mike!

Issue #62, opened 2025-07-31
GLIBC not found -- python module


noajshu:
GLIBC_2.38 not found
I am trying to install tesseract python module in a [colab](https://colab.research.google.com/), and get this error:
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
[/tmp/ipython-input-2523524900.py](https://localhost:8080/#) in <cell line: 0>()
----> 1 import tesseract_decoder
      2 

ImportError: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /usr/local/lib/python3.11/dist-packages/tesseract_decoder.so)

---------------------------------------------------------------------------
NOTE: If your import is failing due to a missing package, you can
manually install dependencies using either !pip or !apt.

To view examples of installing some common dependencies, click the
"Open Examples" button below.
---------------------------------------------------------------------------
```

PR #63, opened 2025-07-31
BUILD on old ubuntu to support colab

NoureldinYosri:
colab has GLIBC 2.35 but building on ubuntu 24 makes us require 2.39 ... building on ubuntu 22 fixes the issue and I also extract the actual GLIBC version and attach it to the pypi wheel

fixes https://github.com/quantumlib/tesseract-decoder/issues/62

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -30,21 +30,18 @@ jobs:
           path: version
 
   build_wheels:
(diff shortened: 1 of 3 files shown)


noajshu (approved):
This is awesome!
I bet that Stim will need a similar fix soon, since I don't think it does anything like this in its workflows.

PR #64, opened 2025-08-01
Arbitrary number of observables

draganaurosgrbic:


src/simplex.h
```diff
@@ -62,14 +63,14 @@ struct SimplexDecoder {
   void decode_to_errors(const std::vector<uint64_t>& detections);
   // Returns the bitwise XOR of all the observables bitmasks of all errors in
   // the predicted errors buffer.
-  common::ObservablesMask mask_from_errors(const std::vector<size_t>& predicted_errors);
+  std::vector<int> mask_from_errors(const std::vector<size_t>& predicted_errors);
```
oscarhiggott: Might be worth renaming this method (including in the pybind) since it's no longer returning a mask

draganaurosgrbic: @oscarhiggott would `get_flipped_observables` (instead of `mask_from_errors`) work?

oscarhiggott: Sounds good to me!

Separate to this but related, I notice in the pybind you're returning a bitstring as `std::vector<bool>` (which is great) but ofc would need a different method name, e.g. `get_observables_from_errors`.

Worth also double checking with @LalehB and @noajshu !

LalehB: `get_observables_from_errors` sounds good to me!

draganaurosgrbic: @LalehB @oscarhiggott got it, thank you!

oscarhiggott: @LalehB I assume you mean for the pybind method and then `get_flipped_observables` for the C++ method here right

src/tesseract.perf.cc
```diff
@@ -33,10 +33,18 @@ void benchmark_decoder(Decoder& decoder, stim::Circuit& circuit, size_t num_shot
   size_t num_low_confidence = 0;
   size_t num_errors = 0;
   size_t num_decoded = 0;
+  auto vector_to_u64_mask = [](const std::vector<int>& v) {
```
LalehB, line 36: the function `vector_to_u64_mask` is defined separately multiple times, you can centralize it to reduce duplication

src/tesseract.cc
```diff
@@ -464,23 +465,36 @@ double TesseractDecoder::cost_from_errors(const std::vector<size_t>& predicted_e
...
+        flipped_observables_set.erase(obs_index);
+      } else {
+        flipped_observables_set.insert(obs_index);
+      }
+    }
   }
-  return mask;
+
+  // Convert the set to a vector and return it.
+  std::vector<int> flipped_observables(flipped_observables_set.begin(),
```
LalehB, line 487: it might be good to sort them before returning

LalehB (approved):
LGTM!
Thanks Dragana, feel free to merge after addressing the comments that Oscar has.

src/tesseract.pybind.h
```diff
@@ -70,15 +70,35 @@ void add_tesseract_module(py::module &root) {
...
            py::arg("detections"), py::arg("det_order"), py::arg("det_beam"))
-      .def("mask_from_errors", &TesseractDecoder::mask_from_errors, py::arg("predicted_errors"))
+      .def(
+          "mask_from_errors",
+          [](TesseractDecoder& self, const std::vector<size_t>& predicted_errors) {
+            std::vector<bool> result(self.num_observables, false);
+            const auto& indices = self.mask_from_errors(predicted_errors);
+            for (int index : indices) {
+              result[index] = true;
+            }
```
oscarhiggott: Rather than getting the indices first (which requires set operations) it might be more efficient to iterate over the errors directly in this routine (and likewise for the analogous simplex method):
```
  for (size_t ei : predicted_errors) {
    for (int obs_index : self.errors[ei].symptom.observables) {
        result[obs_index] ^= true;
    }
  }
```

src/tesseract.pybind.h
```diff
@@ -70,15 +70,35 @@ void add_tesseract_module(py::module &root) {
...
       .def("cost_from_errors", &TesseractDecoder::cost_from_errors, py::arg("predicted_errors"))
-      .def("decode", &TesseractDecoder::decode, py::arg("detections"))
+      .def(
+          "decode",
+          [](TesseractDecoder& self, const std::vector<uint64_t>& detections) {
+            std::vector<bool> result(self.num_observables, false);
+            const auto& indices = self.decode(detections);
+            for (int index : indices) {
+              result[index] = true;
+            }
```
oscarhiggott: Similar to my comment above, it might be more efficient to avoid the set operations since they're not needed here (can do something similar for simplex):

```
  self.decode_to_errors(detections);
  for (size_t ei : self.predicted_errors_buffer) {
    for (int obs_index : self.errors[ei].symptom.observables) {
        result[obs_index] ^= true;
    }
  }
```

src/simplex.h
LalehB: @oscarhiggott yes!

draganaurosgrbic: @LalehB @oscarhiggott Done.

src/tesseract.cc
draganaurosgrbic: @LalehB Done.

src/tesseract.perf.cc
draganaurosgrbic: @LalehB Done.

src/tesseract.pybind.h
draganaurosgrbic: @oscarhiggott Done.

src/tesseract.pybind.h
draganaurosgrbic: @oscarhiggott Done.

PR #65, opened 2025-08-03
Expose `det_order` building in the python module

noajshu:
- The flag `--det-order-bfs` significantly improves the performance of tesseract for decoding e.g. color code circuit-level noise
- There is no way to use it from python currently because the logic is contained in `tesseract_main.cc`
- Here, we move it into utils and expose it in the python module


src/py/utils_test.py
```diff
@@ -44,6 +44,18 @@ def test_build_detector_graph():
...
+def test_build_det_orders():
+    assert tesseract_decoder.utils.build_det_orders(
+        _DETECTOR_ERROR_MODEL, num_det_orders=1, seed=0
+    ) == [[0, 1]]
+
+
+def test_build_det_orders_no_bfs():
+    assert tesseract_decoder.utils.build_det_orders(
+        _DETECTOR_ERROR_MODEL, num_det_orders=1, det_order_bfs=False, seed=0
+    ) == [[0, 1]]
```
LalehB, line 56: It seems that for this DEM, with bfs and no bfs will have the same det_orders, right?

src/utils.cc
```diff
@@ -79,6 +81,91 @@ std::vector<std::vector<size_t>> build_detector_graph(const stim::DetectorErrorM
...
+  std::vector<std::vector<size_t>> det_orders(num_det_orders);
+  std::mt19937_64 rng(seed);
+  std::normal_distribution<double> dist(0, 1);
+
+  auto detector_coords = get_detector_coords(dem);
+
+  if (det_order_bfs) {
+    auto graph = build_detector_graph(dem);
+    std::uniform_int_distribution<size_t> dist_det(0, graph.size() - 1);
+    for (size_t det_order = 0; det_order < num_det_orders; ++det_order) {
```
LalehB, line 96: Thank you so much for moving the `build_det_orders` to the util folder! it makes more sense :-) maybe later we break this `build_det_orders` function into smaller helper functions such as `generate_bfs_order` and more.

LalehB (approved):
LGTM!
Thank you Noah.
btw since the bfs ordering is by default on, is this mostly for the cases that we might want to turn it off?

PR #66, opened 2025-08-05
Accelerating Decoder's Initialization​

draganaurosgrbic:
### Description
This Pull Request introduces a substantial performance optimization to _Tesseract_'s initialization phase. While previous efforts primarily focused on enhancing the critical decoding speed, this work addresses an identified bottleneck in the one-time setup/initialization process. I've targeted a highly inefficient code segment and achieved remarkable speedups.

---

### Background
Before _Tesseract_ can decode simulations/shots of quantum circuits, it must first read and parse the quantum circuit model. This process involves populating and constructing internal data structures essential for decoding. For a given quantum circuit, _Tesseract_ performs this initialization once, then utilizes the constructed data structures and parsed model to decode multiple shots/simulations. As such, the initialization phase hasn't been a primary focus for optimization, as it's a one-time operation and generally not a major time sink compared to the iterative decoding process. However, after achieving significant performance gains in the decoding phase, I identified an opportunity to further improve overall efficiency by optimizing a particularly inefficient loop within initialization.

---

### Problem: Inefficient `eneighbors` Calculation
The primary bottleneck I identified within the initialization phase was the loop responsible for calculating `eneighbors` (error neighbors). This data structure determines, for each error, which detectors are affected by its neighboring errors. The original implementation, shown below, exhibited severe performance issues:
```cpp
  std::vector<std::unordered_set<size_t>> edets_sets(edets.size());
  for (size_t ei = 0; ei < edets.size(); ++ei) {
    edets_sets[ei] = std::unordered_set<size_t>(edets[ei].begin(), edets[ei].end());
  }
  for (size_t ei = 0; ei < num_errors; ++ei) {
    std::set<int> neighbor_set;
    for (int d : edets[ei]) {
      for (int oei : d2e[d]) {
        for (int od : edets_sets[oei]) {
          if (!edets_sets[ei].contains(od)) {
            neighbor_set.insert(od);
          }
        }
      }
    }
    eneighbors[ei] = std::vector<int>(neighbor_set.begin(), neighbor_set.end());
  }
```
This implementation suffered from:
1.  **High Computational Complexity:** The four nested loops resulted in a complexity proportional to `num_errors` \* `detectors_per_error` \* `errors_per_detector` \* `detectors_per_neighbor_error`.
2.  `std::set` and `std::unordered_set` Overheads: Frequent insert operations on `std::set` (logarithmic time complexity) and contains operations on `std::unordered_set` (average constant time complexity) introduced significant memory management overhead and could be substantial when iterated large number of times.

---

### Solution: Leveraging `boost::dynamic_bitset` for Efficient Set Operations
Drawing from the successful application of `boost::dynamic_bitset` in optimizing syndrome pattern hashing (as implemented in #57), I replaced `std::set` and `std::unordered_set` in this critical initialization loop with `boost::dynamic_bitset`. This significantly accelerated the `eneighbors` calculation. As detailed in #57, `boost::dynamic_bitset` offers memory efficiency similar to `std::vector<bool>` but provides highly optimized bit-wise operations for manipulating elements. This is achieved by packing individual bits/elements into contiguous memory blocks and enabling a single bit-wise operation to be executed across multiple elements from the same block simultaneously, leveraging CPU vectorization. The optimized loop is shown below:
```cpp
  std::vector<boost::dynamic_bitset<>> edets_bitsets(num_errors,
                                                     boost::dynamic_bitset<>(num_detectors));
  for (size_t ei = 0; ei < num_errors; ++ei) {
    for (int d : edets[ei]) {
      edets_bitsets[ei][d] = 1;
    }
  }

  for (size_t ei = 0; ei < num_errors; ++ei) {
    boost::dynamic_bitset<> neighbor_set(num_detectors, false);
    for (int d : edets[ei]) {
      for (int oei : d2e[d]) {
        // Unify detectors from neighboring errors
        neighbor_set |= edets_bitsets[oei];
      }
    }
    // Remove detectors from error's own set
    neighbor_set &= ~edets_bitsets[ei];

    for (size_t d = neighbor_set.find_first(); d != boost::dynamic_bitset<>::npos;
         d = neighbor_set.find_next(d)) {
      eneighbors[ei].push_back(d);
    }
  }
```
This optimization significantly improves performance by:
1.  **Reduced Nested Loops:** The code now contains three nested loops (instead of the original four), substantially decreasing the total number of iterations.
2.  **Vectorized Bit-wise Operations:** `boost::dynamic_bitset` stores bits in contiguous memory blocks and enables the execution of highly optimized, hardware-accelerated bit-wise operations. These operations can work on entire blocks in a single CPU cycle, effectively performing vectorized set unions and differences. This dramatically reduces the overhead of element-wise checks and insertions found in the original implementation.
3.  **Memory Efficiency:** `boost::dynamic_bitset` retains the memory-saving bit-packing feature similar to `std::vector<bool>` while eliminating the performance overhead stemming from `std::vector<bool>`'s proxy objects and inefficient bit-level manipulations that operate on individual elements separately.

---

### Impact and Performance Benchmarks
The speedups I achieved in the initialization function are remarkable across various code families and configurations.

#### Before Optimization (Initial Times)
- **Color Codes:** 1.5 - 9 seconds
- **Bivariate-Bicycle Codes:** 6 - 17 seconds
- **Surface Codes:** 0.1 - 0.8 seconds
- **Transversal CNOT Protocols:** 0.2 - 7 seconds

#### After Optimization (Speedups)
- **Color Codes:** 95.89x to 128.18x (less than a second)
- **Bivariate-Bicycle Codes:** 106.43x to 132.88x (less than a second)
- **Surface Codes:** 26.41x to 36.07x (less than a second)
- **Transversal CNOT Protocols:** 17.51x to 43.07x (less than a second)

As shown by the initial times, the initialization previously did not exceed 17 seconds for the benchmarks I performed, with Bivariate-Bicycle codes having the highest overhead. However, since this operation is performed once per quantum circuit (and Tesseract then uses the initialized knowledge to decode multiple simulations/shots, where performance is critical), even these initial times were acceptable.

Nevertheless, initialization times after this optimization fell below a second for all tested code families and configurations. For Color Codes, initialization fell below 0.09 seconds, for Bivariate-Bicycle Codes below 0.15 seconds, For Surface Codes below 0.03 seconds and for Transversal CNOT Protocols below 0.4 seconds. This dramatic reduction explains the exceptionally high speedup factors; **the initialization phase is now extremely fast.**

Below are plots that show the performance gains I achieved across different code families and configurations.

color1

color2

color3

bicycle1

bicycle2

bicycle3

surface1

trans1

trans2

trans3

---

### Conclusion
This optimization to the initialization function demonstrates the substantial performance gains achievable by refactoring inefficient loops and leveraging advanced data structures like `boost::dynamic_bitset`. It showcases how its highly optimized bit-wise operations (enabling vectorized execution across multiple elements at once) can be used to implement highly efficient set operations (union, difference). The resulting remarkable speedups further enhance the overall efficiency of the Tesseract decoder.

---

### Key Contributions
* **Identified and Investigated:** Pinpointed a critical inefficiency/loop within the initialization function that was consuming significant time.
* **Leveraged Advanced Data Structures:** Applied knowledge and success from optimization in #57 to replace code logic that frequently manipulates set elements. This involved leveraging `boost::dynamic_bitset` for highly optimized, vectorized bit-wise operations to perform set operations.
* **Refactored Critical Loop:** Replaced the inefficient `eneighbors` calculation loop, which previously used `std::unordered_set` and `std::set`, with an improved version utilizing `boost::dynamic_bitset` and vectorized bit-wise operations for set unions and differences.
* **Achieved Remarkable Speedups:** Delivered exceptional speedups across various code families and configurations, reaching up to 132.88x faster initialization in a Bivariate-Bicycle code benchmark, making Tesseract even more robust and scalable.

### DIFF
--- a/src/simplex.pybind.h
+++ b/src/simplex.pybind.h
@@ -15,6 +15,7 @@
 #ifndef _SIMPLEX_PYBIND_H
 #define _SIMPLEX_PYBIND_H
 
+#include <pybind11/iostream.h>
 #include <pybind11/operators.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
@@ -63,7 +64,8 @@ void add_simplex_module(py::module& root) {
       .def_readwrite("end_time_to_errors", &SimplexDecoder::end_time_to_errors)
       .def_readonly("low_confidence_flag", &SimplexDecoder::low_confidence_flag)
       .def("init_ilp", &SimplexDecoder::init_ilp)
-      .def("decode_to_errors", &SimplexDecoder::decode_to_errors, py::arg("detections"))
+      .def("decode_to_errors", &SimplexDecoder::decode_to_errors, py::arg("detections"),
+           py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>())
       .def(
           "get_observables_from_errors",
           [](SimplexDecoder& self, const std::vector<size_t>& predicted_errors) {
@@ -88,6 +90,7 @@ void add_simplex_module(py::module& root) {
             }
             return result;
           },
-          py::arg("detections"));
+          py::arg("detections"),
+          py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>());
 }
 #endif
--- a/src/tesseract.pybind.cc
+++ b/src/tesseract.pybind.cc
@@ -14,6 +14,7 @@
 
 #include "tesseract.pybind.h"
 
+#include <pybind11/iostream.h>
 #include <pybind11/pybind11.h>
 
 #include "common.pybind.h"
@@ -23,8 +24,18 @@
 
 PYBIND11_MODULE(tesseract_decoder, tesseract) {
   py::module::import("stim");
+
   add_common_module(tesseract);
   add_utils_module(tesseract);
   add_simplex_module(tesseract);
   add_tesseract_module(tesseract);
+
+  // Adds a context manager to the python library that can be used to redirect C++'s stdout/stderr
+  // to python's stdout/stderr at run time like
+  // with tesseract_decoder.ostream_redirect(stdout=..., stderr=...):
+  //    do_work()
+  // This is only needed if the C++ function's stdout/stderr is not redirected to python's
+  // stdout/stderr using the py::call_guard<py::scoped_ostream_redirect,
+  // py::scoped_estream_redirect>() statement.
+  py::add_ostream_redirect(tesseract, "ostream_redirect");
 }
--- a/src/tesseract.pybind.h
+++ b/src/tesseract.pybind.h
@@ -15,6 +15,7 @@
 #ifndef _TESSERACT_PYBIND_H
 #define _TESSERACT_PYBIND_H
 
+#include <pybind11/iostream.h>
 #include <pybind11/operators.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
@@ -71,11 +72,13 @@ void add_tesseract_module(py::module& root) {
       .def(py::init<TesseractConfig>(), py::arg("config"))
       .def("decode_to_errors",
            py::overload_cast<const std::vector<uint64_t>&>(&TesseractDecoder::decode_to_errors),
-           py::arg("detections"))
+           py::arg("detections"),
+           py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>())
       .def("decode_to_errors",
            py::overload_cast<const std::vector<uint64_t>&, size_t, size_t>(
                &TesseractDecoder::decode_to_errors),
-           py::arg("detections"), py::arg("det_order"), py::arg("det_beam"))
+           py::arg("detections"), py::arg("det_order"), py::arg("det_beam"),
+           py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>())
       .def(
           "get_observables_from_errors",
           [](TesseractDecoder& self, const std::vector<size_t>& predicted_errors) {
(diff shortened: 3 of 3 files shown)


src/tesseract.cc
```diff
@@ -148,22 +148,32 @@ void TesseractDecoder::initialize_structures(size_t num_detectors) {
...
-    std::set<int> neighbor_set;
+    boost::dynamic_bitset<> neighbor_set(num_detectors, false);
     for (int d : edets[ei]) {
       for (int oei : d2e[d]) {
-        for (int od : edets[oei]) {
-          if (!edets_sets[ei].contains(od)) {
-            neighbor_set.insert(od);
-          }
-        }
+        if (oei == ei) continue;
```
LalehB: do we need this line since self detectors are removed later on with this line `neighbor_set &= ~edets_bitsets[ei];`

src/tesseract.cc
```diff
@@ -148,22 +148,32 @@ void TesseractDecoder::initialize_structures(size_t num_detectors) {
...
+        // Unify detectors from neighboring errors
+        neighbor_set |= edets_bitsets[oei];
+      }
+    }
+    // Remove detectors from error's own set
+    neighbor_set &= ~edets_bitsets[ei];
+
+    for (size_t d = 0; d < num_detectors; ++d) {
```
LalehB: suggestion: since this is an iteration over the dynamic_bitset of `neighbor_set` you can avoid iterating over all the detectors by using `find_first()/find_next()`

LalehB (commented):
Thanks Dragana!
I added a few comments, but could you please also add some tests for correctness (i.e., checking the accuracy)?

src/tesseract.cc
draganaurosgrbic: @LalehB Done.

src/tesseract.cc
draganaurosgrbic: @LalehB Done.

draganaurosgrbic:
@LalehB I have added the tests that compare/check the `eneighbors` calculation inside `Tesseract`.

PR #67, opened 2025-08-06
Update verbose logging mechanism to enable callback provided by user

noajshu:
Refactored verbose logging mechanism to use a `CallbackStream` class that can be given a callback handler
Since AI agents have trouble with Bazel, I added a CMake build system.

- If the Python users leaves the `verbose_callback=None` (the default) in the config ctor, it will be a no-op and all verbose logging will be disabled.
- This should enable integration with the visualization utilities in viz/, which could probably be moved into python now and integrated better with the module.


NoureldinYosri:
did you run the clang formatter on the code locally? ... the GitHub clang formatter can't parse the code

src/simplex.cc
```diff
@@ -197,9 +213,7 @@ void SimplexDecoder::decode_to_errors(const std::vector<uint64_t>& detections) {
         add_costs_for_time(t1);
         ++t1;
       }
-      if (config.verbose) {
-        std::cout << "t0 = " << t0 << " t1 = " << t1 << std::endl;
-      }
+      config.log_stream << "t0 = " << t0 << " t1 = " << t1 << std::endl;
```
LalehB, line 216: should we also have this: `if (config.log_stream.active) {` for this log_stream?

CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
+cmake_minimum_required(VERSION 3.16)
```
NoureldinYosri, line 1: this change is unrelated to the functionality in the PR title

noajshu:
Abandoning this because https://github.com/quantumlib/tesseract-decoder/pull/68 fixes the problem

PR #71, opened 2025-08-07
ignoring the unrecognized commands

LalehB:
This PR just replaces asserts to ignore the commands that are not important to the decoder

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -126,7 +126,7 @@ stim::DetectorErrorModel common::remove_zero_probability_errors(
         out_dem.append_dem_instruction(instruction);
         break;
       default:
-        std::cout << "Unrecognized instruction type: " << instruction.type << std::endl;
+        throw std::invalid_argument("Unrecognized instruction type: " + op.str());
     }
   }
   return out_dem;
--- a/src/common.cc
+++ b/src/common.cc
@@ -79,7 +79,7 @@ stim::DetectorErrorModel common::merge_identical_errors(const stim::DetectorErro
   for (const stim::DemInstruction& instruction : dem.flattened().instructions) {
     switch (instruction.type) {
       case stim::DemInstructionType::DEM_SHIFT_DETECTORS:
-        assert(false && "unreachable");
+        std::cout << "Unrecognized instruction: DEM_SHIFT_DETECTORS" << std::endl;
         break;
       case stim::DemInstructionType::DEM_ERROR: {
         Error error(instruction);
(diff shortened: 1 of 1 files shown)


src/common.cc
```diff
@@ -79,7 +79,7 @@ stim::DetectorErrorModel common::merge_identical_errors(const stim::DetectorErro
   for (const stim::DemInstruction& instruction : dem.flattened().instructions) {
     switch (instruction.type) {
       case stim::DemInstructionType::DEM_SHIFT_DETECTORS:
-        assert(false && "unreachable");
+        std::cout << "Unrecognized instruction: DEM_SHIFT_DETECTORS" << std::endl;
         break;
```
oscarhiggott: Can this case ever occur? Since the dem is flattened (`dem.flattened().instructions`) I don't think there should be any `stim::DemInstructionType::DEM_SHIFT_DETECTORS` or `stim::DemInstructionType::DEM_REPEAT_BLOCK` instructions.

src/common.cc
```diff
@@ -126,7 +126,7 @@ stim::DetectorErrorModel common::remove_zero_probability_errors(
         out_dem.append_dem_instruction(instruction);
         break;
       default:
-        assert(false && "unreachable");
+        std::cout << "Unrecognized instruction type: " << instruction.type << std::endl;
```
oscarhiggott: Once all the known `stim::DemInstructionType` are handled, the default could throw an exception (which pybind will convert into a ValueError).

```suggestion
        throw std::invalid_argument("Unrecognized instruction type: " + op.str());
```

With this approach there should be a separate case for each of these:
```
enum class DemInstructionType : uint8_t {
    DEM_ERROR,
    DEM_SHIFT_DETECTORS,
    DEM_DETECTOR,
    DEM_LOGICAL_OBSERVABLE,
    DEM_REPEAT_BLOCK,
};
```

of which all but `DEM_ERROR` and `DEM_DETECTOR` can be silently ignored (no need for `std::cout`).

oscarhiggott (approved):
Looks good but added some suggestions in the comments

src/common.cc
```diff
@@ -79,7 +79,7 @@ stim::DetectorErrorModel common::merge_identical_errors(const stim::DetectorErro
   for (const stim::DemInstruction& instruction : dem.flattened().instructions) {
     switch (instruction.type) {
       case stim::DemInstructionType::DEM_SHIFT_DETECTORS:
-        assert(false && "unreachable");
+        std::cout << "Unrecognized instruction: DEM_SHIFT_DETECTORS" << std::endl;
```
NoureldinYosri: maybe `cerr`?

src/common.cc
LalehB: yeah I thought if the flattening had some issue/bug and this InstructionType showed up it would be nice to catch it, but I am okay with removing it totally too!

src/common.cc
LalehB: sure! done.

PR #72, opened 2025-08-07
Keeps the `stim::DemInstructionType::DEM_LOGICAL_OBSERVABLE` in the dem during processing

oscarhiggott:
This ensures that the number of logical observables from `dem.count_observables()` remains correct, even if there `DEM_LOGICAL_OBSERVABLE` observables not triggered by an error.

I also had it throw an exception for unrecognized instructions as we really should be handling all the cases correctly now (and it would be good for the user to know if we're not). Can switch back to cerr if necessary though.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -99,7 +99,7 @@ stim::DetectorErrorModel common::merge_identical_errors(const stim::DetectorErro
         break;
       }
       default:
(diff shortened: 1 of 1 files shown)


LalehB (approved):
LGTM!

Issue #70, opened 2025-08-07
Make `tesseract.decode()` function signature consistent with stim detector sampler


oscarhiggott:
Since most people will probably use tesseract with stim, I think it would be nice if the decode method input/output more closely aligned with that of [`stim.CompiledDetectorSampler.sample()`](https://github.com/quantumlib/Stim/blob/main/doc/python_api_reference_vDev.md#stim.CompiledDetectorSampler.sample).

For example, rather than taking as input a list of detection event indices, it could instead take as input a numpy array of dtype bool and return a numpy array of dtype bool.

For example this might involve changing this function:
```
      .def(
          "decode",
          [](TesseractDecoder& self, const std::vector<uint64_t>& detections) {
            std::vector<bool> result(self.num_observables, false);
            self.decode(detections);
            for (size_t ei : self.predicted_errors_buffer) {
              for (int obs_index : self.errors[ei].symptom.observables) {
                result[obs_index] = result[obs_index] ^ true;
              }
            }
            return result;
          },
```
to the following (untested!) function:
```
      .def(
          "decode",
          [](TesseractDecoder& self, const py::array_t<bool>& syndrome) {
            std::vector<size_t> detections;
            auto syndrome_unchecked = syndrome.unchecked<1>();
            for (size_t i = 0; i < syndrome_unchecked.size(); i++) {
              if (syndrome_unchecked(i)) {
                detections.push_back(i);
              }
            }
            self.decode(detections);
            std::vector<bool> result(self.num_observables, false);
            for (size_t ei : self.predicted_errors_buffer) {
              for (int obs_index : self.errors[ei].symptom.observables) {
                result[obs_index] = result[obs_index] ^ true;
              }
            }
            return py::array(result.size(), result.data());
          },
```

and similarly for simplex. Of course this isn't backward compatible, but maybe that's ok since we're currently in a dev version.

draganaurosgrbic:
Resolved in #73

PR #73, opened 2025-08-07
Python Interface: Decode from detection events and decode batches of shots

draganaurosgrbic:
Address issue #70

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,11 +65,5 @@ jobs:
       run: |
         brew install clang-format
 
-    - name: Check formatting (redundant but allowed)
-      run: |
-        clang-format --version
-        files=$(git ls-files '*.cc' '*.cpp' '*.h' '*.hpp')
-        [ -z "$files" ] || clang-format --dry-run --Werror $files
-
     - name: Bazel tests
       run: bazel test src/...
--- a/src/py/simplex_test.py
+++ b/src/py/simplex_test.py
@@ -18,13 +18,14 @@ import stim
 import numpy as np
 
 from src import tesseract_decoder
-from src.py.test_utils import (
-    shared_test_decode_from_detection_events,
+from src.py.shared_decoding_tests import (
+    shared_test_decode,
     shared_test_decode_batch_with_invalid_dimensions,
     shared_test_decode_batch_with_complex_model,
     shared_test_decoder_predicts_various_observable_flips,
-    shared_test_decode_from_detection_events_complex_dem,
+    shared_test_decode_complex_dem,
     shared_test_decode_batch,
+    shared_test_decode_from_detection_events,
 )
 
 _DETECTOR_ERROR_MODEL = stim.DetectorErrorModel(
@@ -66,16 +67,16 @@ def test_simplex_decoder_predicts_various_observable_flips(decoder_class, config
     "decoder_class, config_class",
     [(tesseract_decoder.simplex.SimplexDecoder, tesseract_decoder.simplex.SimplexConfig)]
 )
-def test_simplex_decode_from_detection_events(decoder_class, config_class):
-    shared_test_decode_from_detection_events(decoder_class, config_class)
+def test_simplex_decode(decoder_class, config_class):
+    shared_test_decode(decoder_class, config_class)
 
 
 @pytest.mark.parametrize(
     "decoder_class, config_class",
     [(tesseract_decoder.simplex.SimplexDecoder, tesseract_decoder.simplex.SimplexConfig)]
 )
-def test_simplex_decode_from_detection_events_complex_dem(decoder_class, config_class):
-    shared_test_decode_from_detection_events_complex_dem(decoder_class, config_class)
+def test_simplex_decode_complex_dem(decoder_class, config_class):
+    shared_test_decode_complex_dem(decoder_class, config_class)
 
 
 @pytest.mark.parametrize(
@@ -102,6 +103,12 @@ def test_simplex_decode_batch_with_complex_model(decoder_class, config_class):
     shared_test_decode_batch_with_complex_model(decoder_class, config_class)
 
 
+@pytest.mark.parametrize(
+    "decoder_class, config_class",
+    [(tesseract_decoder.simplex.SimplexDecoder, tesseract_decoder.simplex.SimplexConfig)]
+)
+def test_simplex_decode_from_detection_events(decoder_class, config_class):
+    shared_test_decode_from_detection_events(decoder_class, config_class)
 
 if __name__ == "__main__":
     raise SystemExit(pytest.main([__file__]))
--- a/src/py/simplex_test.py
+++ b/src/py/simplex_test.py
@@ -62,7 +62,7 @@ def test_simplex_decoder_predicts_various_observable_flips():
         dem = stim.DetectorErrorModel(dem_string)
 
         # Initialize SimplexConfig and SimplexDecoder with the generated DEM
-        config = tesseract_decoder.simplex.SimplexConfig(dem, window_length=1)
+        config = tesseract_decoder.simplex.SimplexConfig(dem)
         decoder = tesseract_decoder.simplex.SimplexDecoder(config)
 
         # Simulate a detection event on D0.
@@ -96,7 +96,7 @@ def test_decode_from_detection_events():
     dem = stim.DetectorErrorModel(dem_string)
 
     # Configure the decoder.
-    config = tesseract_decoder.simplex.SimplexConfig(dem, window_length=1)
+    config = tesseract_decoder.simplex.SimplexConfig(dem)
     decoder = tesseract_decoder.simplex.SimplexDecoder(config)
     decoder.init_ilp()
 
@@ -140,7 +140,7 @@ def test_decode_from_detection_events_complex_dem():
     dem = stim.DetectorErrorModel(dem_string)
 
     # Configure the decoder.
-    config = tesseract_decoder.simplex.SimplexConfig(dem, window_length=1)
+    config = tesseract_decoder.simplex.SimplexConfig(dem)
     decoder = tesseract_decoder.simplex.SimplexDecoder(config)
     decoder.init_ilp()
 
@@ -168,7 +168,7 @@ def test_decode_batch_with_invalid_dimensions():
     dem = stim.DetectorErrorModel(dem_string)
 
     # Configure the decoder.
-    config = tesseract_decoder.simplex.SimplexConfig(dem, window_length=1)
+    config = tesseract_decoder.simplex.SimplexConfig(dem)
     decoder = tesseract_decoder.simplex.SimplexDecoder(config)
     decoder.init_ilp()
 
@@ -190,7 +190,7 @@ def test_decode_batch():
         error(0.05) D1
     '''
     dem = stim.DetectorErrorModel(dem_string)
-    config = tesseract_decoder.simplex.SimplexConfig(dem, window_length=1)
+    config = tesseract_decoder.simplex.SimplexConfig(dem)
     decoder = tesseract_decoder.simplex.SimplexDecoder(config)
     decoder.init_ilp()
 
@@ -235,7 +235,7 @@ def test_decode_batch_more_complex_dem():
     dem = stim.DetectorErrorModel(dem_string)
 
     # Configure the decoder.
-    config = tesseract_decoder.simplex.SimplexConfig(dem, window_length=1)
+    config = tesseract_decoder.simplex.SimplexConfig(dem)
     decoder = tesseract_decoder.simplex.SimplexDecoder(config)
     decoder.init_ilp()
 
--- a/src/py/simplex_test.py
+++ b/src/py/simplex_test.py
@@ -18,6 +18,14 @@ import stim
 import numpy as np
 
 from src import tesseract_decoder
+from src.py.test_utils import (
+    shared_test_decode_from_detection_events,
+    shared_test_decode_batch_with_invalid_dimensions,
+    shared_test_decode_batch_with_complex_model,
+    shared_test_decoder_predicts_various_observable_flips,
+    shared_test_decode_from_detection_events_complex_dem,
+    shared_test_decode_batch,
+)
 
 _DETECTOR_ERROR_MODEL = stim.DetectorErrorModel(
     """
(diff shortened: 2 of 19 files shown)


src/tesseract.pybind.h
```diff
@@ -83,28 +84,89 @@ void add_tesseract_module(py::module& root) {
...
+            }
+
             return result;
           },
-          py::arg("detections"),
-          py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>())
+          py::arg("syndromes"),
+          "Decodes a batch of shots (a 2D NumPy array of booleans) and "
+          "returns a 2D NumPy array of predicted observable flips, where "
+          "each row corresponds to a shot.")
```
oscarhiggott: It might be nice to provide a full docstring, e.g. in numpydoc format, with arguments and return type documented

draganaurosgrbic:
@LalehB @noajshu I merged my PR with the main branch, but it looks like the main branch had version control conflict markers inside source files. I fixed this on my branch, but the main branch still has them.

src/simplex.cc
```diff
@@ -238,27 +238,13 @@ void SimplexDecoder::decode_to_errors(const std::vector<uint64_t>& detections) {
       if (config.verbose) {
         // Get the solution information
         const HighsInfo& info = highs->getInfo();
-<<<<<<< HEAD
-        config.log_stream << "Simplex iteration count: " << info.simplex_iteration_count
-                          << std::endl;
-        config.log_stream << "Objective function value: " << info.objective_function_value
```
draganaurosgrbic, line 244: This error came from the main branch, not from my work. I merged my PR with the main branch, and the main branch had these conflict markers left inside. I fixed this on my branch and once we merge my PR, main will be fine as well.

src/simplex.cc
```diff
@@ -303,25 +289,13 @@ void SimplexDecoder::decode_to_errors(const std::vector<uint64_t>& detections) {
     if (config.verbose) {
       // Get the solution information
       const HighsInfo& info = highs->getInfo();
-<<<<<<< HEAD
-      config.log_stream << "Simplex iteration count: " << info.simplex_iteration_count << std::endl;
-      config.log_stream << "Objective function value: " << info.objective_function_value
-                        << std::endl;
```
draganaurosgrbic, line 309: This error came from the main branch, not from my work. I merged my PR with the main branch, and the main branch had these conflict markers left inside. I fixed this on my branch and once we merge my PR, main will be fine as well.

.github/workflows/ci.yml
```diff
@@ -65,11 +65,5 @@ jobs:
       run: |
         brew install clang-format
 
-    - name: Check formatting (redundant but allowed)
-      run: |
-        clang-format --version
-        files=$(git ls-files '*.cc' '*.cpp' '*.h' '*.hpp')
-        [ -z "$files" ] || clang-format --dry-run --Werror $files
-
```
noajshu, line 73: curious why we remove formatting check?

src/py/simplex_test.py
```diff
@@ -44,18 +45,15 @@ def test_create_simplex_decoder():
     decoder.decode_to_errors([1])
     assert decoder.get_observables_from_errors([1]) == []
     assert decoder.cost_from_errors([2]) == pytest.approx(1.0986123)
-    assert decoder.decode([1]) == []
```
noajshu, line 47: why remove the actual decoding step?

src/py/simplex_test.py
```diff
@@ -84,5 +83,186 @@ def test_simplex_decoder_predicts_various_observable_flips():
...
+    """
+    dem_string = f'''
+        error(0.1) D0 D1 L0
+    '''
+
+    # Create the DEM.
+    dem = stim.DetectorErrorModel(dem_string)
+
+    # Configure the decoder.
+    config = tesseract_decoder.simplex.SimplexConfig(dem, window_length=1)
```
noajshu: Use of the `window_length` argument is really not recommended unless you're specifically decoding a long memory experiment and want to use sliding window decoding.

src/py/tesseract_test.py
```diff
@@ -50,18 +51,17 @@ def test_create_decoder():
     decoder.decode_to_errors(detections=[0], det_order=0, det_beam=0)
     assert decoder.get_observables_from_errors([1]) == []
     assert decoder.cost_from_errors([1]) == pytest.approx(0.5108256237659907)
-    assert decoder.decode([0]) == []
```
noajshu, line 53: Ditto -- could we keep exercising the decoding part?

src/py/tesseract_test.py
```diff
@@ -76,20 +76,204 @@ def test_tesseract_decoder_predicts_various_observable_flips():
...
+
+    # Expected batch predictions based on the DEM.
+    expected_predictions = np.array([
+        [True, False, False],   # Shot 1: L0 flip
+        [False, True, False],   # Shot 2: L1 flip
+        [True, False, True],    # Shot 3: L0 and L2 flips
+        [False, False, False],  # Shot 4: No flips
+    ], dtype=bool)
+
+    assert np.array_equal(actual_predictions, expected_predictions)
```
noajshu: Optional nit : these tests are a bit repetitive and maybe could be refactored to use the same examples / logic for both tesseract and simplex? WDYT?

src/simplex.cc
noajshu: Sorry about that! That came from an accident on my own fork accidentally getting merged back to this repo 😨 
Thanks for catching it!!

src/tesseract.pybind.h
draganaurosgrbic: @oscarhiggott Thank you. Done.

src/py/simplex_test.py
draganaurosgrbic: It is not removed, it is renamed. The wrapper does not have a "decode" function anymore, we replaced it with "decode_from_detection_events", and now I have separate tests just for that function.

.github/workflows/ci.yml
draganaurosgrbic: This was the fix for the pipeline. There are two phases inside the CI: clang-format and bazel tests. The bazel tests (macos and ubuntu) run also the clang-format (again). The problem is the phase that performs the clang-format itself uses one version (version 18), and the one inside macos uses another version (version 20, as it is installed with `brew install`). This will make it impossible for me to pass the pipeline for this PR, as one version was requesting one formatting, another different. This issue will stay persistent, I brought this up before, but until this PR, there were no conflicts between those two versions. When I submitted this PR, it was failing because one phase was requesting one formatting, and another different formatting.

src/py/tesseract_test.py
draganaurosgrbic: See above, this function is renamed with "decode_from_detection_events" and I made separate tests for that function.

src/py/simplex_test.py
draganaurosgrbic: Thank you! I removed it.

src/py/simplex_test.py
```diff
@@ -84,5 +83,186 @@ def test_simplex_decoder_predicts_various_observable_flips():
...
+    # Create a simple DEM with two detectors and one observable.
+    dem_string = f'''
+        error(0.1) D0 D1 L0
+        error(0.2) D0
+    '''
+    dem = stim.DetectorErrorModel(dem_string)
+
+    # Configure the decoder.
+    config = tesseract_decoder.simplex.SimplexConfig(dem)
+    decoder = tesseract_decoder.simplex.SimplexDecoder(config)
```
oscarhiggott: What do we think about adding a method `tesseract_decoder.simplex.SimplexConfig.compile_decoder()` or similar that outputs the `tesseract_decoder.simplex.SimplexDecoder`?

Then can just do:
`decoder = tesseract_decoder.simplex.SimplexConfig(dem).compile_decoder()`

Not necessary for this PR so could make it an issue if it's of interest (and likewise for tesseract)

src/simplex.pybind.h
```diff
@@ -70,27 +71,116 @@ void add_simplex_module(py::module& root) {
...
+            auto syndrome_unchecked = syndrome.unchecked<1>();
+            for (size_t i = 0; i < syndrome_unchecked.size(); i++) {
+              if (syndrome_unchecked(i)) {
+                detections.push_back(i);
+              }
+            }
             self.decode(detections);
+            // Note: `std::vector<bool>` is a special C++ template that does not
+            // provide a contiguous memory block, which is required by `pybind11`
+            // for direct NumPy array creation. Therefose, I use `std::vector<char>`
```
oscarhiggott:
```suggestion
            // for direct NumPy array creation. Therefore, I use `std::vector<char>`
```

src/simplex.pybind.h
```diff
@@ -70,27 +71,116 @@ void add_simplex_module(py::module& root) {
...
             }
             return result;
           },
           py::arg("predicted_errors"))
       .def("cost_from_errors", &SimplexDecoder::cost_from_errors, py::arg("predicted_errors"))
       .def(
-          "decode",
-          [](SimplexDecoder& self, const std::vector<uint64_t>& detections) {
-            std::vector<bool> result(self.num_observables, false);
+          "decode_from_detection_events",
```
oscarhiggott, line 84: My proposal was that the `decode` method takes a bitstring of detector outcomes as input (i.e. this function) but that we add a new method `decode_from_detection_events` which is implemented how `decode` _used_ to be defined.

```suggestion
          "decode",
```

In other words, `decode` should have arguments `(SimplexDecoder& self, const py::array_t<bool>& syndrome)` and `decode_from_detection_events` should have arguments `(SimplexDecoder& self, const std::vector<uint64_t>& detections)`

src/py/tesseract_test.py
draganaurosgrbic: @noajshu Thank you! Done.

src/tesseract.pybind.h
```diff
@@ -83,28 +84,116 @@ void add_tesseract_module(py::module& root) {
...
             return result;
           },
           py::arg("predicted_errors"))
       .def("cost_from_errors", &TesseractDecoder::cost_from_errors, py::arg("predicted_errors"))
       .def(
-          "decode",
-          [](TesseractDecoder& self, const std::vector<uint64_t>& detections) {
-            std::vector<bool> result(self.num_observables, false);
+          "decode_from_detection_events",
+          [](TesseractDecoder& self, const py::array_t<bool>& syndrome) {
```
oscarhiggott: As mentioned above for simplex, this method (the one that takes detector measurements as input) should be called `decode`, and there should be a new method `decode_from_detection_events` that has the original function signature (i.e. it takes detection events as input, not the bitstring). See my above comment on simplex for more details.

```suggestion
          "decode",
          [](TesseractDecoder& self, const py::array_t<bool>& syndrome) {
```

src/tesseract.pybind.h
```diff
@@ -83,28 +84,116 @@ void add_tesseract_module(py::module& root) {
...
+          },
+          py::arg("syndrome"),
+          py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>(),
+          R"pbdoc(
+        Decodes a single shot.
+
+        Parameters
+        ----------
+        syndrome : np.ndarray
+            A 1D NumPy array of booleans representing the detection events for a single shot.
```
oscarhiggott:
```suggestion
            A 1D NumPy array of booleans representing the detector outcomes for a single shot.
```

src/tesseract.pybind.h
```diff
@@ -83,28 +84,116 @@ void add_tesseract_module(py::module& root) {
...
+                  detections.push_back(j);
+                }
+              }
+              self.decode(detections);
+
+              // Collect results for the current shot being decoded.
+              std::vector<char> shot_result(self.num_observables, 0);
+              for (size_t ei : self.predicted_errors_buffer) {
+                for (int obs_index : self.errors[ei].symptom.observables) {
+                  shot_result[obs_index] ^= true;
```
oscarhiggott, line 167: It might be simpler to XOR straight into the numpy array, for example how about something like the following:

```suggestion
                  result_unchecked(i, obs_index) ^= true;
```

src/tesseract.pybind.h
```diff
@@ -83,28 +84,116 @@ void add_tesseract_module(py::module& root) {
...
+                }
+              }
+
+              // Copy the result into the pre-allocated array.
+              for (size_t k = 0; k < self.num_observables; ++k) {
+                result_unchecked(i, k) = shot_result[k];
+              }
```
oscarhiggott: Rather than allocating a new vector and copying, it might be simpler to do something like this:

```suggestion
              for (size_t ei : self.predicted_errors_buffer) {
                for (int obs_index : self.errors[ei].symptom.observables) {
                  result_unchecked(i, obs_index) ^= true;
                }
              }
```

src/simplex.pybind.h
```diff
@@ -70,27 +71,116 @@ void add_simplex_module(py::module& root) {
...
+              for (size_t ei : self.predicted_errors_buffer) {
+                for (int obs_index : self.errors[ei].symptom.observables) {
+                  shot_result[obs_index] ^= 1;
+                }
+              }
+
+              // Copy the result into the pre-allocated array.
+              for (size_t k = 0; k < self.num_observables; ++k) {
+                result_unchecked(i, k) = shot_result[k];
+              }
```
oscarhiggott: Similar to my comment below, it might be simpler to do something like this which avoids the copying:

```suggestion
              for (size_t ei : self.predicted_errors_buffer) {
                for (int obs_index : self.errors[ei].symptom.observables) {
                  result_unchecked(i, obs_index) ^= 1;
                }
              }
```

src/simplex.pybind.h
```diff
@@ -70,27 +71,116 @@ void add_simplex_module(py::module& root) {
...
-          py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>());
+          py::arg("syndromes"),
+          R"pbdoc(
+        Decodes a batch of shots.
+
+        Parameters
+        ----------
+        syndromes : np.ndarray
+            A 2D NumPy array of booleans where each row represents a single shot's
+            detection events. The shape should be (num_shots, num_detectors): each shot has
```
oscarhiggott:
```suggestion
            detector outcomes. The shape should be (num_shots, num_detectors): each shot has
```

src/tesseract.pybind.h
draganaurosgrbic: @oscarhiggott Thank you. Done.

src/simplex.pybind.h
draganaurosgrbic: @oscarhiggott Thank you. Done.

src/tesseract.pybind.h
draganaurosgrbic: @oscarhiggott Thank you. Done.

src/simplex.pybind.h
draganaurosgrbic: @oscarhiggott Thank you. Done.

src/py/simplex_test.py
draganaurosgrbic: @oscarhiggott Fine with me, for another issue. This PR has, I think, already enough content.

oscarhiggott: Ok just added issue #75 for this

draganaurosgrbic: @oscarhiggott  Thank you. Can I merge this one?

src/simplex.pybind.h
```diff
@@ -70,27 +71,145 @@ void add_simplex_module(py::module& root) {
...
           py::arg("detections"),
-          py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>());
+          py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>(),
+          R"pbdoc(
+          Decodes a single shot from a list of detection events.
+
+          Parameters
+          ----------
+          detections : list[int]
+              A list of `uint64_t` indices corresponding to the detectors that were
```
oscarhiggott: Maybe remove this uint64_t type since the docstring is viewed in python
```suggestion
              A list of indices corresponding to the detectors that were
```

src/simplex.pybind.h
```diff
@@ -70,27 +71,145 @@ void add_simplex_module(py::module& root) {
...
+        syndromes : np.ndarray
+            A 2D NumPy array of booleans where each row represents a single shot's
+            detector outcomes. The shape should be (num_shots, num_detectors): each shot has
+            a new array with num_detectors size.
+
+        Returns
+        -------
+        np.ndarray
+            A 2D NumPy array of booleans where each row corresponds to a shot and
+            that short specifies which logical observable are flipped. The shape is
```
oscarhiggott:
```suggestion
            A 2D NumPy array of booleans where each row corresponds to a shot and
            each column corresponds to a logical observable. Each row is the decoder's prediction of which observables were flipped in the shot. The shape is
```

src/tesseract.pybind.h
```diff
@@ -83,28 +84,145 @@ void add_tesseract_module(py::module& root) {
...
+          },
+          py::arg("syndromes"),
+          R"pbdoc(
+        Decodes a batch of shots.
+
+        Parameters
+        ----------
+        syndromes : np.ndarray
+            A 2D NumPy array of booleans where each row represents a single shot's
+            detection events. The shape should be (num_shots, num_detectors): each shot has
```
oscarhiggott:
```suggestion
            A 2D NumPy array of booleans where each row corresponds to a shot and
            each column corresponds to a logical observable. Each row is the decoder's prediction of which observables were flipped in the shot. The shape is
```

src/py/simplex_test.py
oscarhiggott: Added a couple more comments but otherwise looks excellent! Yes feel free to merge once these are addressed

src/tesseract.pybind.h
```diff
@@ -83,28 +84,145 @@ void add_tesseract_module(py::module& root) {
...
           py::arg("detections"),
-          py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>())
+          py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>(),
+          R"pbdoc(
+          Decodes a single shot from a list of detection events.
+
+          Parameters
+          ----------
+          detections : list[int]
+              A list of `uint64_t` indices corresponding to the detectors that were
```
oscarhiggott:
```suggestion
              A list of indices corresponding to the detectors that were
```

src/tesseract.pybind.h
```diff
@@ -83,28 +84,145 @@ void add_tesseract_module(py::module& root) {
...
+              fired. This input represents a single measurement shot.
+
+          Returns
+          -------
+          np.ndarray
+              A 1D NumPy array of booleans. Each boolean value indicates whether the
+              corresponding logical observable has been flipped by the decoded error.
```
oscarhiggott:
```suggestion
              decoder predicts that the corresponding logical observable has been flipped.
```

src/py/simplex_test.py
draganaurosgrbic: @oscarhiggott Thank you so much!

PR #74, opened 2025-08-07
Add tesseract-decoder-dev email address to README file

mhucka:
Now that the team has an email address for direct contact, we can use that instead of the generic quantum-oss-maintainers@google.com list as the contact address.


LalehB (approved):
LGTM!
thank you Mike :-)

PR #77, opened 2025-08-08
Update documentation for Python interface

draganaurosgrbic:


LalehB (commented):
Thank you @draganaurosgrbic !
I have a few suggestions:
* The bullet list of methods for tesseract in the “Explanation of each method” section still lists the old decode(detections: list[int]) API; consider updating it to include decode_from_detection_events, decode, and decode_batch for consistency

* Align parameter names in the decode_to_errors overload with the implementation (det_order/det_beam rather than detector_order/detector_beam) to avoid confusion when referencing the code

* The example code that follows the new section still demonstrates decoding via decode(detections); consider showing an example using the new decode_from_detection_events or decode (syndrome array) APIs for completeness

LalehB (approved):
LGTM!
thank you Dragana!

Issue #54, opened 2025-07-16
add docstrings and typehints to the python interface


NoureldinYosri:
this can be done through pypbind11 itself https://pybind11.readthedocs.io/en/stable/advanced/misc.html or by creating stub files

draganaurosgrbic:
Resolved in #80

PR #80, opened 2025-08-08
Add missing docstrings and tests inside Python interface

draganaurosgrbic:
Address issue #54

I have added docstrings and tests where they were missing for each class we have in Python Interface. Also, inside the `shared_decoding_tests` module that I created, I have moved tests that are the same for `Simplex` and `Tesseract`, so we have it more generic.

Our Python interface/wrapper should look much prettier now.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -109,12 +109,10 @@ stim::DetectorErrorModel common::merge_indistinguishable_errors(
         if (error.symptom.detectors.size() == 0) {
           throw std::invalid_argument("Errors that do not flip any detectors are not supported.");
         }
-        error.likelihood_cost =
-            -1 * std::log(instruction.arg_data[0] / (1 - instruction.arg_data[0]));
 
         if (errors_by_symptom.find(error.symptom) != errors_by_symptom.end()) {
-          double existing_cost = errors_by_symptom[error.symptom].likelihood_cost;
-          error.likelihood_cost = merge_weights(existing_cost, error.likelihood_cost);
+          error.likelihood_cost = merge_weights(error.likelihood_cost,
+                                                errors_by_symptom[error.symptom].likelihood_cost);
         }
         errors_by_symptom[error.symptom] = error;
         break;
--- a/src/common.test.cc
+++ b/src/common.test.cc
@@ -127,3 +127,57 @@ TEST(CommonTest, merge_weights_is_equivalent_to_probability_xor) {
   // Test with identical probabilities
   assert_merged_probabilities_are_equal(0.01, 0.01);
 }
+
+// Helper function to create a simple DEM with two identical errors.
+stim::DetectorErrorModel create_dem_with_two_errors(double p1, double p2) {
+  stim::DetectorErrorModel dem;
+
+  std::vector<stim::DemTarget> targets = {stim::DemTarget::relative_detector_id(0)};
+
+  dem.append_error_instruction(p1, targets, "");
+  dem.append_error_instruction(p2, targets, "");
+  return dem;
+}
+
+// Function to get the probability from a merged DEM.
+double get_merged_probability(const stim::DetectorErrorModel& merged_dem) {
+  if (merged_dem.instructions.size() != 1 ||
+      merged_dem.instructions[0].type != stim::DemInstructionType::DEM_ERROR) {
+    throw std::runtime_error("Expected a single DEM_ERROR instruction in the merged model.");
+  }
+  return merged_dem.instructions[0].arg_data[0];
+}
+
+TEST(CommonTest, merge_indistinguishable_errors_two_errors) {
+  // Case 1: Both probabilities are low.
+  double p1 = 0.1;
+  double p2 = 0.2;
+  double expected_merged_p = p1 * (1 - p2) + p2 * (1 - p1);
+  auto dem1 = create_dem_with_two_errors(p1, p2);
+  auto merged_dem1 = common::merge_indistinguishable_errors(dem1);
+  ASSERT_NEAR(get_merged_probability(merged_dem1), expected_merged_p, 1e-9);
+
+  // Case 2: One low, one high probability.
+  p1 = 0.1;
+  p2 = 0.8;
+  expected_merged_p = p1 * (1 - p2) + p2 * (1 - p1);
+  auto dem2 = create_dem_with_two_errors(p1, p2);
+  auto merged_dem2 = common::merge_indistinguishable_errors(dem2);
+  ASSERT_NEAR(get_merged_probability(merged_dem2), expected_merged_p, 1e-9);
+
+  // Case 3: One high, one low probability.
+  p1 = 0.8;
+  p2 = 0.1;
+  expected_merged_p = p1 * (1 - p2) + p2 * (1 - p1);
+  auto dem3 = create_dem_with_two_errors(p1, p2);
+  auto merged_dem3 = common::merge_indistinguishable_errors(dem3);
+  ASSERT_NEAR(get_merged_probability(merged_dem3), expected_merged_p, 1e-9);
+
+  // Case 4: Both probabilities are high.
+  p1 = 0.8;
+  p2 = 0.9;
+  expected_merged_p = p1 * (1 - p2) + p2 * (1 - p1);
+  auto dem4 = create_dem_with_two_errors(p1, p2);
+  auto merged_dem4 = common::merge_indistinguishable_errors(dem4);
+  ASSERT_NEAR(get_merged_probability(merged_dem4), expected_merged_p, 1e-9);
+}


src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
-      .def_readwrite("detectors", &common::Symptom::detectors)
-      .def_readwrite("observables", &common::Symptom::observables)
+           py::arg("detectors") = std::vector<int>(), py::arg("observables") = std::vector<int>(),
+           R"pbdoc(
+            The constructor for the `Symptom` class.
+
+            Parameters
+            ----------
+            detectors : list[int], default=[]
+                The indices of the detectors affected by this symptom.
```
oscarhiggott:
```suggestion
                The indices of the detectors in this symptom.
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
 void add_common_module(py::module &root) {
   auto m = root.def_submodule("common", "classes commonly used by the decoder");
 
-  py::class_<common::Symptom>(m, "Symptom")
+  py::class_<common::Symptom>(m, "Symptom", R"pbdoc(
+        Represents a syndrome in terms of detectors and observables.
```
oscarhiggott:
```suggestion
        Represents a symptom of an error, which is a list of detectors and a list of observables
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+
+        A symptom is defined by a list of detectors it affects and a list of
```
oscarhiggott:
```suggestion
        A symptom is defined by a list of detectors that are flipped and a list of
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+        observables it flips.
```
oscarhiggott:
```suggestion
        observables that are flipped.
```

src/common.pybind.h
oscarhiggott:
```suggestion
                The indices of the flipped detectors.
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+           py::arg("detectors") = std::vector<int>(), py::arg("observables") = std::vector<int>(),
+           R"pbdoc(
+            The constructor for the `Symptom` class.
+
+            Parameters
+            ----------
+            detectors : list[int], default=[]
+                The indices of the detectors affected by this symptom.
+            observables : list[int], default=[]
+                The indices of the observables flipped by this symptom.
```
oscarhiggott:
```suggestion
                The indices of the flipped observables.
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+           )pbdoc")
+      .def_readwrite("detectors", &common::Symptom::detectors,
+                     "A list of detector indices that are flipped by this symptom.")
```
oscarhiggott:
```suggestion
                     "A list of the detector indices that are flipped in this symptom.")
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+      .def_readwrite("observables", &common::Symptom::observables,
+                     "A list of observable indices that are flipped by this symptom.")
```
oscarhiggott:
```suggestion
                     "A list of observable indices that are flipped in this symptom.")
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+        Returns
+        -------
+        list[stim.DemTarget]
+            A list of `stim.DemTarget` objects representing the detectors and observables.
+      )pbdoc");
+
+  py::class_<common::Error>(m, "Error", R"pbdoc(
+        Represents an error, including its cost, probability, and symptom.
+
+        An error links a physical event to the logical syndrome it produces.
```
oscarhiggott:
```suggestion
        An error is a physical event (or set of indistinguishable physical events)
        defined by the detectors and observables that it flips in the circuit.
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+    )pbdoc")
+      .def_readwrite("likelihood_cost", &common::Error::likelihood_cost,
+                     "The cost of this error (often -log(probability)).")
```
oscarhiggott: Does tesseract use -log(p)?

Shouldn't this be log((1-p)/p)?

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
       .def(py::init<double, std::vector<int> &, std::vector<int>, std::vector<bool> &>(),
            py::arg("likelihood_cost"), py::arg("detectors"), py::arg("observables"),
-           py::arg("dets_array"))
+           py::arg("dets_array"), R"pbdoc(
+            Constructor for the `Error` class.
+
+            Parameters
+            ----------
+            likelihood_cost : float
+                The cost of this error. This is often `-log(probability)`.
```
oscarhiggott: This should be:
```suggestion
                The cost of this error. 
                This is often `log((1 - probability) / probability)`.
```

We should check this is always what's used in the codebase

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+            detectors : list[int]
+                A list of detector indices affected by this error.
```
oscarhiggott:
```suggestion
                A list of indices of the detectors flipped by this error.
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+            observables : list[int]
+                A list of observable indices flipped by this error.
```
oscarhiggott:
```suggestion
                A list of indices of the observables flipped by this error.
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+            dets_array : list[bool]
+                A boolean array representing the detectors affected.
```
oscarhiggott: How is this dets_array actually used?

If it's not used shall we remove it from the codebase? It seems like a different representation of the `detectors` argument already provided, so seems like duplication.

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
       .def(py::init<double, double, std::vector<int> &, std::vector<int>, std::vector<bool> &>(),
            py::arg("likelihood_cost"), py::arg("probability"), py::arg("detectors"),
-           py::arg("observables"), py::arg("dets_array"))
+           py::arg("observables"), py::arg("dets_array"), R"pbdoc(
+            Constructor with both likelihood cost and probability.
+
+            Parameters
+            ----------
+            likelihood_cost : float
+                The cost of this error.
```
oscarhiggott:
```suggestion
                The cost of this error, usually log((1 - probability) / probability)
```

src/common.pybind.h
```diff
@@ -29,40 +29,110 @@ namespace py = pybind11;
...
+            probability : float
```
oscarhiggott: The `likelihood_cost` and `probability` arguments are closely related, since one is usually a function of the other (e.g. log((1-p)/p)). 

It might be helpful to provide documentation on how they are both used in tesseract. If you provide both likelihood_cost and probability, which one is used by the decoder when decoding? @noajshu @LalehB any thoughts?

src/common.pybind.h
```diff
@@ -71,23 +141,66 @@ void add_common_module(py::module &root) {
         auto res = common::merge_identical_errors(input_dem);
         return make_py_object(res, "DetectorErrorModel");
       },
-      py::arg("dem"));
+      py::arg("dem"), R"pbdoc(
+        Merges identical errors in a `stim.DetectorErrorModel`.
```
oscarhiggott, line 149:
```suggestion
        Merges identical errors in a `stim.DetectorErrorModel`.
        
        Errors are identical if they flip the same set of detectors and observables (the same symptom).
        For example, two identical errors with probabilities p1 and p2
        would be merged into a single error with the same symptom,
        but with probability `p1 * (1 - p2) + p2 * (1 - p1)`
```

src/common.pybind.h
draganaurosgrbic: I just checked, it's not used neither in C++ nor in the Python wrapper. Should I remove it from the Python wrapper only or also from C++?

draganaurosgrbic:
> I think it would be good to rename `common::merge_identical_errors` to `common::merge_indistinguishable_errors` and change "identical" to "indistinguishable" in the corresponding docstrings.
> 
> Identical makes it sound like the errors have the same probability as well, when really they just have the same symptom.

@oscarhiggott Thank you. Done.

src/common.pybind.h
draganaurosgrbic: `likelihood_cost` is used when decoding. It is used when computing the heuristic cost. And initially, it is calculated using the `probability`.

src/common.pybind.h
draganaurosgrbic: I checked, yes it is always `log((1 - probability) / probability)`, my mistake with the initial formula.

src/common.pybind.h
draganaurosgrbic: I checked, yes it is always `log((1 - probability) / probability)`, my mistake with the initial formula.

src/common.pybind.h
oscarhiggott: So what happens when a user provides both likelihood_cost and probability? Which one is used for the likelihood cost?

src/common.pybind.h
oscarhiggott: I would remove it from both but maybe double check with @LalehB @noajshu

src/common.pybind.h
draganaurosgrbic: I can see in C++ there are separate constructors that receive only `likelihood_cost` and that receive both `likelihood_cost` and `probability`. If both `likelihood_cost` and `probability` are provided, then they are directly set to these values. But, these seem like generic constructors that were written here, I don't see these constructors actually being called inside the code. Instead, the main constructor that we actually call in the code (from our decoders) is the one that constructors the error from `stim::DemInstruction`. This main constructor will extract the `probability` from `stim::DemInstruction` and then also use it to calculate the `likelihood_cost` (using the formula `log((1-p)/p)`).

src/common.pybind.h
draganaurosgrbic: @LalehB @noajshu it looks like we don't use `dets_array`. Should I remove it from the Python wrapper only or from C++ as well?

src/common.pybind.h
draganaurosgrbic: I just tested removing these default/generic constructors (that receive all arguments separately), both from C++ and Python, and compiling passes. It looks like they are not used in the code at all. Maybe I could completely remove them? We only construct `Error` from `stim::DemInstruction`.

oscarhiggott: Might it be that there just isn't enough test coverage of the non-dem case?

draganaurosgrbic: This would also remove the `dets_array` we don't use anywhere.

draganaurosgrbic: We can have tests that only test these constructors, but these constructors are never called from a decoder. They just sit there and they are not called by any other code. We never construct errors using these constructors.

draganaurosgrbic: So from `Tesseract` and `Simplex`, we only construct errors using `stim::DemInstruction`. We can keep these default constructors, but I am just saying we never use them when we start decoding with `Tesseract` or `Simplex`.

draganaurosgrbic: @LalehB @noajshu What are your thoughts on this? We have two constructors inside the `Error` class that receive all arguments the class has, but we never use them/call them by `Tesseract` or `Simplex`. We only construct `Error` objects using `std::DemInstruction` objects. Should I remove them? We are also exposing these constructors in the Python wrapper, but when we decode with `Tesseract` and `Simplex`, we always construct errors from `std::DemInstruction`.

oscarhiggott: What I _think_ is probably the best thing to do for now, is to keep only 
* One constructor that has signature `py::init<double, std::vector<int> &, std::vector<int>` (`likelihood_cost`, `detectors`, `observables`)
* The constructor that loads from a DEM error

oscarhiggott: Both of these can be in the python API and C++

draganaurosgrbic: Sounds good to me. And I can add a test in Python that tests the first constructor (since it looks like it was not tested anywhere).

draganaurosgrbic: And we remove `dets_array` from `Error`. That one is really not used anywhere.

oscarhiggott: I could be wrong, but I think the only place that `probability` is actually being used is for merging errors [here](https://github.com/quantumlib/tesseract-decoder/blob/6da2b8f2eb2c377facaeb7249cc5f96f5f5cc1ff/src/common.cc#L87).

But the likelihood costs can be merged directly using [this function](https://github.com/oscarhiggott/PyMatching/blob/04e956fcd3da97f9ac2613df7794a7ac764c4bba/src/pymatching/sparse_blossom/driver/io.cc#L19-L23) (feel free to copy-paste). So once that change is made I think the `probability` data member can be removed entirely?

draganaurosgrbic: I think you are right! Then we wouldn't have to store `probability` as the class property. I will try this.

oscarhiggott: Somehow github didn't load some of your more recent messages before you replied. 

Is it plausible that in the future the decoders could be configured directly from a list of `Error` instances? I guess the config explicitly has a dem as an attribute, so maybe such a change wouldn't be backward compatible anyway. Anyway I don't see much harm in keeping the two constructors we discussed, but agree that currently only the dem one is used.

draganaurosgrbic: Agree! Then I will first see if we can remove `probability` (and use the formula you sent me), and then have these two constructors we discussed about.

oscarhiggott: I do worry that the current API depends a bit too heavily on a DEM. I think a better structure of the code would be if the configs were constructed _from_ a dem, but then only stored a list of these `Error` instances internally.

That way, the configs could be loaded from other file formats, e.g. by a user specifying each `Error` themself, or by loading from a check matrix or stabilizer code directly.

draganaurosgrbic: @oscarhiggott Your idea with removing `probability` worked! I also removed `dets_array` and left the two constructors we discussed. Our `Error` class now has only `likelihood_cost` and `symptom` properties. I also have a test that verifies that the formula you sent me was computing the same merge logic as we did when we had explicit probability fields.

src/common.pybind.h
draganaurosgrbic: Resolved. Removed `dets_array` from the `Error` class.

src/common.pybind.h
draganaurosgrbic: @LalehB I also removed these fields from the README.md.

src/common.cc
```diff
@@ -26,7 +26,7 @@ std::string common::Symptom::str() {
 
 common::Error::Error(const stim::DemInstruction& error) {
   assert(error.type == stim::DemInstructionType::DEM_ERROR);
```
oscarhiggott, line 32:
```suggestion
  if (error.type != stim::DemInstructionType::DEM_ERROR) {
     throw std::invalid_argument("Error must be loaded from an error dem instruction, but received: " + error.str());
  }
  assert(error.type == stim::DemInstructionType::DEM_ERROR);
```

src/common.cc
```diff
@@ -26,7 +26,7 @@ std::string common::Symptom::str() {
...
-  probability = error.arg_data[0];
+  double probability = error.arg_data[0];
   assert(probability >= 0 && probability <= 1);
```
oscarhiggott:
```suggestion
if (probability < 0 || probability > 1) {
    throw std::invalid_argument("Probability must be between 0 and 1, but received: " + std::to_string(probability));
}
```

src/common.cc
```diff
@@ -81,12 +89,13 @@ stim::DetectorErrorModel common::merge_identical_errors(const stim::DetectorErro
       case stim::DemInstructionType::DEM_ERROR: {
         Error error(instruction);
         assert(error.symptom.detectors.size());
```
oscarhiggott: Really we should handle this case correctly, but in the meantime we should at least ensure we don't crash the python interpreter:

```suggestion
        if (error.symptom.detectors.size() == 0) {
            throw std::invalid_argument("Errors that do not flip any detectors are not supported.");
        }
```

src/common.cc
```diff
@@ -81,12 +89,13 @@ stim::DetectorErrorModel common::merge_identical_errors(const stim::DetectorErro
...
-        // Merge with existing error with the same symptom (if applicable)
+        error.likelihood_cost =
+            -1 * std::log(instruction.arg_data[0] / (1 - instruction.arg_data[0]));
+
         if (errors_by_symptom.find(error.symptom) != errors_by_symptom.end()) {
-          double p0 = errors_by_symptom[error.symptom].probability;
-          error.probability = p0 * (1 - error.probability) + (1 - p0) * error.probability;
+          double existing_cost = errors_by_symptom[error.symptom].likelihood_cost;
+          error.likelihood_cost = -merge_weights(-existing_cost, -error.likelihood_cost);
```
oscarhiggott: I would expect the following to work, so if it doesn't we should understand why:

```suggestion
          error.likelihood_cost = merge_weights(existing_cost, error.likelihood_cost);
```

draganaurosgrbic: Yes, works.

src/common.h
```diff
@@ -80,6 +71,8 @@ stim::DetectorErrorModel remove_zero_probability_errors(const stim::DetectorErro
 stim::DetectorErrorModel dem_from_counts(stim::DetectorErrorModel& orig_dem,
                                          const std::vector<size_t>& error_counts, size_t num_shots);
 
+double merge_weights(double a, double b);
```
oscarhiggott, line 84: The formula for this merge_weights is not so intuitive, so I recommend adding a docstring similar to [the one in pymatching](https://github.com/oscarhiggott/PyMatching/blob/04e956fcd3da97f9ac2613df7794a7ac764c4bba/src/pymatching/sparse_blossom/driver/io.h#L25-L41). You could add a link to the corresponding pymatching function instead/as well.

src/common.test.cc
```diff
@@ -95,3 +95,35 @@ TEST(common, RemoveZeroProbabilityErrors) {
...
+
+// Helper function to compare the two methods.
+void assert_merged_probabilities_are_equal(double p1, double p2) {
+  // Method 1: Merge probabilities directly using the exclusive OR formula.
+  double merged_p_direct = p1 + p2 - 2 * p1 * p2;
+
+  // Method 2: Convert to likelihood costs, merge them, then convert back.
+  double cost1 = -1 * std::log(p1 / (1 - p1));
+  double cost2 = -1 * std::log(p2 / (1 - p2));
+  double merged_cost = common::merge_weights(-cost1, -cost2);
```
oscarhiggott: There is no reason to flip the sign here. The merge_weights function takes as input two likelihood costs just as defined in the code here. The weight in the context you would have seen this function is just `log((1-p)/p)` which is identical to the `-log(p/(1-p))` (how it's implemented in the tesseract code).

```suggestion
  double merged_cost = common::merge_weights(cost1, cost2);
```

src/common.h
draganaurosgrbic: Thank you. Done.

src/common.test.cc
draganaurosgrbic: Thank you. Done.

src/common.cc
oscarhiggott: I'm actually quite concerned that the current version (before my suggested change) is passing the tests. Can we check if there is a test for the `common::merge_indistinguishable_errors` method for a DEM that has two identical errors with different probabilities?

E.g. that a dem:
```
error(p1) D0
error(p2) D0
```

gets merged into a dem:
```
error(p1 * (1 - p2) + p2 * (1 - p1)) D0
```
for four combinations of p1 and p2 where each is both positive and negative.

src/tesseract.pybind.h
```diff
@@ -205,8 +334,8 @@ void add_tesseract_module(py::module& root) {
               }
               self.decode(detections);
 
-              // Note: I must do this if I want to modify the results on the 'result_unchecked'
-              // itself.
+              // Note: I must do this if I want to modify the results
+              // on the 'result_unchecked' itself.
               for (size_t k = 0; k < self.num_observables; ++k) {
                 result_unchecked(i, k) = 0;
               }
```
oscarhiggott: Nit: would it be simpler/more efficient to use `result.attr("fill")(0);` a few lines earlier and remove this for loop, e.g. replace lines 324-5 with:

```
py::array_t<bool> result({num_shots, self.num_observables});
result.attr("fill")(0);
auto result_unchecked = result.mutable_unchecked<2>();
```

src/utils.pybind.h
```diff
@@ -27,36 +27,102 @@ void add_utils_module(py::module &root) {
...
+        ----------
+        dem : stim.DetectorErrorModel
+            The detector error model used to build the graph.
+
+        Returns
+        -------
+        list[list[int]]
+            An adjacency list representation of the detector graph.
+            Each inner list contains the indices of detectors connected
+            to the detector at the corresponding index.
```
oscarhiggott, line 76:
```suggestion
            to the detector at the corresponding index.
            Here we say that two detectors are connected if there exists at
            least one error in the DEM which flips both detectors.
```

src/common.cc
```diff
@@ -103,8 +121,9 @@ stim::DetectorErrorModel common::merge_identical_errors(const stim::DetectorErro
     }
   }
   for (const auto& it : errors_by_symptom) {
-    out_dem.append_error_instruction(it.second.probability,
-                                     it.second.symptom.as_dem_instruction_targets(),
+    // Recalculate probability from the final merged cost for the output DEM.
+    double final_p = 1 / (1 + std::exp(it.second.likelihood_cost));
```
oscarhiggott: Nit: it might be nice to have a `get_probability` method (also exposed in pybind) which calculates the probability from the likelihood_cost, which could also be used here.

It's good that we now only store the likelihood cost, but still could be handy to have a getter/setter that uses probability as helper methods.

src/tesseract.pybind.h
```diff
@@ -107,8 +208,36 @@ void add_tesseract_module(py::module& root) {
...
+            Returns
+            -------
+            list[bool]
+                A list of booleans, where each boolean corresponds to a
+                logical observable and is `True` if the observable was flipped.
+           )pbdoc")
+      .def("cost_from_errors", &TesseractDecoder::cost_from_errors, py::arg("predicted_errors"),
+           R"pbdoc(
+            Calculates the total logarithmic probability cost for a given set of
+            predicted errors.
```
oscarhiggott:
```suggestion
            Calculates the sum of the likelihood costs of the predicted errors.
            The likelihood cost of an error with probability p is log((1 - p) / p).
```

src/tesseract.pybind.h
```diff
@@ -107,8 +208,36 @@ void add_tesseract_module(py::module& root) {
...
+
+            Parameters
+            ----------
+            predicted_errors : list[int]
+                A list of integers representing the predicted error indices.
+
+            Returns
+            -------
+            float
+                A float representing the total logarithmic probability cost.
```
oscarhiggott:
```suggestion
                A float representing the sum of the likelihood costs of the
                predicted errors.
```

src/simplex.pybind.h
```diff
@@ -192,8 +287,8 @@ void add_simplex_module(py::module& root) {
               }
               self.decode(detections);
 
-              // Note: I must do this if I want to modify the results on the 'result_unchecked'
-              // itself.
+              // Note: I must do this if I want to modify the results
+              // on the 'result_unchecked' itself.
               for (size_t k = 0; k < self.num_observables; ++k) {
                 result_unchecked(i, k) = 0;
               }
```
oscarhiggott: See my comment above about using:

```
py::array_t<bool> result({num_shots, self.num_observables});
result.attr("fill")(0);
auto result_unchecked = result.mutable_unchecked<2>();
```

draganaurosgrbic: Thank you. Done.

src/common.cc
draganaurosgrbic: Thank you. Done.

src/tesseract.pybind.h
draganaurosgrbic: Thank you. Done.

src/common.cc
draganaurosgrbic: If I understood what you want, I have added the tests. Check `TEST(CommonTest, merge_indistinguishable_errors_two_errors)`, please.

oscarhiggott: Yes exactly, looks great!

oscarhiggott (approved):
This turned into a pretty epic PR, looks great! Thanks for addressing all my comments.

PR #83, opened 2025-08-09
Create a visualization class and wire it through the code

NoureldinYosri:
Currently visualization is created by parsing the logs ... this PR introduces a different way to do this by creating a visualization class ... for now this class can only write to file to be inline with how things are done now but can be improved in different ways ... for example it can record data that the python layer can use directly with matplotlib

to use the class set the flag `create_visualization=True` in the TesseractConfig and after running the decoding call `decoder.visualizer.write(file_path)` to write to the desired file


src/visualization.cc
```diff
@@ -0,0 +1,53 @@
...
+  for (size_t oei : activated_errors) {
+    ss << oei << ", ";
+  }
+  lines.push_back(ss.str());
+}
+
+void Visualizer::add_activated_detectors(const boost::dynamic_bitset<>& detectors,
+                                         size_t num_detectors) {
+  std::stringstream ss;
+  ss << "activated_detectors = ";
```
NoureldinYosri, line 35: @LalehB is it supposed to be activated_detectors or activated_dets ... the script seems to look for _dets https://github.com/quantumlib/tesseract-decoder/blob/0a7aa241f0f96fba905986d8ee47abc8e500f880/viz/to_json.py#L49

LalehB: @NoureldinYosri yeah amazing catch! it used to be `activated_dets` and then in the code got changed to `activated_detectors` but apparently the visualization did not get updated! thanks a lot!

LalehB (approved):
LGTM!

PR #88, opened 2025-08-11
Make Tesseract compatible with Sinter

draganaurosgrbic:


src/tesseract_sinter_compat.pybind.cc
```diff
@@ -0,0 +1,292 @@
...
+  bool operator!=(const TesseractSinterDecoder& other) const {
+    return !(*this == other);
+  }
+
+  // Take a string representation of the DEM, parse the DEM and return a compiled decoder instance.
+  TesseractSinterCompiledDecoder compile_decoder_for_dem(const py::object& dem) {
+    const stim::DetectorErrorModel stim_dem(py::cast<std::string>(py::str(dem)).c_str());
+
+    TesseractConfig config;
+    config.dem = stim_dem;
```
oscarhiggott: It would be nice if the various hyperparameters of tesseract were exposed in the sinter custom decoder class (I don't think this is currently done in your implementation?). Perhaps the constructor for the sinter decoder could expose these as arguments and use them in the compiled decoder and decode via files? Or maybe the sinter decoder can just accept a config itself as an optional argument?

src/BUILD
```diff
@@ -91,12 +91,20 @@ pybind_extension(
     ],
 )
 
+pybind_extension(
+    name = "tesseract_sinter_compat",
+    srcs = ["tesseract_sinter_compat.pybind.cc"],
+    deps = [
+        ":libtesseract",
+    ],
+)
```
oscarhiggott: Is there a reason for adding this as a separate library, rather than just adding it into the tesseract_decoder library?

src/tesseract_sinter_compat.pybind.cc
draganaurosgrbic: Thank you. Done.

src/BUILD
draganaurosgrbic: I realize now your suggestion is better. Thank you. Done.

draganaurosgrbic:
@oscarhiggott Thank you so much for the serialization reference! I have updated the code and the Sinter Integration decoder is now also serializable. I also added a section about the Sinter Integration inside the README.md for the Python interface.

src/py/README.md
```diff
@@ -481,3 +481,91 @@ print("\nEstimated DEM:")
...
+#### The TesseractSinterDecoder Object
+All Sinter examples rely on this utility function to provide the Sinter-compatible Tesseract decoder.
+
+```python
+import sinter
+import stim
+from sinter._decoding._decoding import sample_decode
+
+from src.tesseract_decoder import tesseract_sinter_compat as tesseract_module
+from src import tesseract_decoder
```
oscarhiggott, line 497: It would be better if this code could be directly used by someone who pip installed the package. E.g.:
```suggestion
from tesseract_decoder import tesseract_sinter_compat as tesseract_module
import tesseract_decoder
```

However as mentioned in my other comment, this example code doesn't currently work for me.

draganaurosgrbic:
@oscarhiggott the test passes in the CI. As I mentioned before, you need to run tests with `bazel test src/...`

src/py/README.md
draganaurosgrbic: @oscarhiggott You need to run tests with `bazel test src/...`

draganaurosgrbic: @oscarhiggott I have not set the configuration of the project, the configuration of the project requires that all tests are run with `bazel test src/...`. All tests use the import syntax that you see above, not just this one

oscarhiggott:
> @oscarhiggott I had added that exactly same test that you told me that fails in the CI. It passed when I ran it. Do you have an idea why it failed when you run it?? Did you pull my latest changes? You can wait for the CI to complete and see if that test fails in the CI.

Hi @draganaurosgrbic, I'm not sure what the bug is, I haven't had the chance to take a look into it. But the tests pass fine. There's something that the tests don't catch that still causes sinter not to work. I would recommend trying to reproduce the issue I found using the method I gave to install the wheel. The tests pass for me too (using the latest version) but the sinter does not work with the code snippet I showed.

oscarhiggott:
Though actually running the same code with pymatching also fails so maybe there's a bug in the example itself

oscarhiggott:
Would be good if the examples were updated to work (be in `if __name__ == "__main__"`)

oscarhiggott:
> > Would be good if the examples were updated to work (be in `if __name__ == "__main__"`). Would also be nice if the python readme was just directly in the top level docs directory (e.g. `python_documentation.md` or similar)
> 
> 
> 
> I didn't make a decision where to place the Python documentation, but I can move it, if you find it more appropriate

I do think having it in docs as docs/python_documentation.md or similar would be best as users will look for documentation in the docs folder first

oscarhiggott:
Good point another PR is best to keep the diff small, can wait until this is merged. I'll take another look through this PR now.

draganaurosgrbic:
> Good point another PR is best to keep the diff small, can wait until this is merged. I'll take another look through this PR now.

Thank you so much!!!

oscarhiggott:
One nice thing to have would be a test along these lines (this is a sketch, would need to replace config_arg1 etc). Do you think that's doable in this PR or shall we create an issue for it?

```

@pytest.mark.parametrize("config_arg_1,config_arg_2, etc", [("ex1_arg1", "ex1_arg2"), ("ex2_arg1", "ex2_arg2")])
def test_decode_shots_bit_packed_vs_decode_batch(config_arg_1, config_arg2, etc):
    circuit = stim.Circuit.generated(
        "color_code:memory_xyz",
        distance=3,
        rounds=3,
        after_clifford_depolarization=0.02
    )
    dem = circuit.detector_error_model()
    
    config = tesseract_decoder.tesseract.TesseractConfig(config_arg_1, config_arg2)

    sinter_decoder = tesseract_module.TesseractSinterDecoder(config=config)
    compiled_sinter_decoder = sinter_decoder.compile_decoder_for_dem(dem=dem)
    
    decoder = config.compile_decoder()
    
    bitpacked_shots, obs = circuit.compile_detector_sampler().sample(shots=1000, separate_observables=True, bit_packed=True)
    
    unpacked_shots = np.unpackbits(bitpacked_shots, bitorder='little', axis=1)
    
    predictions_sinter = compiled_sinter_decoder.decode_shots_bit_packed(
        bit_packed_detection_event_data=bitpacked_shots)
    
    predictions_decode_batch = decoder.decode_batch(unpacked_shots)
    
    assert np.array_equal(predictions_sinter, predictions_decode_batch)

```

draganaurosgrbic:
@oscarhiggott Please, another issue.

draganaurosgrbic:
@oscarhiggott Please, check my last commit. It has a new test.

src/tesseract_sinter_compat.pybind.h
```diff
@@ -0,0 +1,332 @@
...
+  bool operator!=(const TesseractSinterDecoder& other) const {
+    return !(*this == other);
+  }
+
+  // Take a string representation of the DEM, parse the DEM and return a compiled decoder instance.
+  TesseractSinterCompiledDecoder compile_decoder_for_dem(const py::object& dem) {
+    const stim::DetectorErrorModel stim_dem(py::cast<std::string>(py::str(dem)).c_str());
+
+    TesseractConfig local_config = config;
+    local_config.dem = stim_dem;
```
oscarhiggott, line 125: It would be better if the first argument of the `TesseractConfig` (at least in the python bindings) was optional and defaulted to None (or some other default).

At the moment, constructing a custom sinter decoder requires adding this dummy DEM, which is a bit ugly:

```
import stim
from tesseract_decoder.tesseract_sinter_compat import TesseractSinterDecoder
from tesseract_decoder.tesseract import TesseractConfig

config = TesseractConfig(dem=stim.DetectorErrorModel(), det_beam=15)
sinter_decoder = TesseractSinterDecoder(config=config)
```

I guess there could then also be a `TesseractConfig.compile_decoder_for_dem(dem: stim.DetectorErrorModel) -> TesseractDecoder` method (as well as the current `TesseractConfig.compile_decoder()` method) for using the standard decoding class.

draganaurosgrbic:
@oscarhiggott please write me all the issues you have today, so I can address them by tomorrow (tomorrow is my last day). If you have anything else you want me to change, please add by today.

src/tesseract_sinter_compat.pybind.h
draganaurosgrbic: Thank you. Done.

PR #93, opened 2025-08-12
Add observables to `tesseract_decoder.common.Error.__str__`

oscarhiggott:


draganaurosgrbic:
@oscarhiggott I fixed the tests. The CI is passing now.

noajshu (approved):

src/common.cc
```diff
@@ -81,7 +82,9 @@ common::Error::Error(const stim::DemInstruction& error) {
 }
 
 std::string common::Error::str() const {
-  return "Error{cost=" + std::to_string(likelihood_cost) + ", symptom=" + symptom.str() + "}";
+  std::stringstream ss;
+  ss << std::fixed << std::setprecision(2) << likelihood_cost;
```
oscarhiggott, line 86: @draganaurosgrbic any chance we can revert this `std::setprecision(2)`? I think just `std::to_string(likelihood_cost)` is best so that a user can see exactly what the error is being assumed by the decoder.

draganaurosgrbic: I think we should fix the number of decimals when creating a string representation. It doesn't have to be 2, it can be more (e.g. 6 decimals). Not just because we are testing the function (although we could write a different test that validates whether the string matches a pattern), but I think we should fix the number of decimals, to make printing statements across different errors consistent. Maybe use 6 decimals?

draganaurosgrbic: @oscarhiggott The default number of decimals when printing `double` values in C++ is 6. Therefore, it will never be more than 6. I can fix the precision to 6, if you like that.

oscarhiggott: Ahh I see, I think 6 would be better if you're happy to change that, thanks!

draganaurosgrbic: @oscarhiggott Please see #96

PR #100, opened 2025-08-12
Add issue templates

mhucka:
This adds templates based on other Quantumlib projects. This follows the practices of other projects in providing a more guided workflow for people to report bugs, ask questions, or take note of tasks to be done.

To see an example of what these templates end up looking like once they're installed in a repository, click on the <kbd>New issue</kbd> button at https://github.com/quantumlib/qsim/issues

### DIFF
--- a/.github/ISSUE_TEMPLATE/3-task.yml
+++ b/.github/ISSUE_TEMPLATE/3-task.yml
@@ -1,0 +1,34 @@
+name: Task
+description: Describe a task that needs to be done
+type: Task
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for taking the time to fill out this feature request form!
+
+  - type: textarea
+    attributes:
+      label: What is the task?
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/3-task.yml
```diff
@@ -0,0 +1,34 @@
+name: Task
+description: Describe a task that needs to be done
+type: Task
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for taking the time to fill out this feature request form!
```
LalehB: > WDYT about maybe changing the "feature request" to "reference task" to make it differentiate with the previous form (feature request)

Oh, yes, good point. Will do in a commit in just a moment. Thanks for catching that.

LalehB (approved):
LGTM!
thank you Mike, this is pretty nice!

PR #101, opened 2025-08-12
Add workflow to label stale issues and PRs

mhucka:
As more and more issues and pull requests start to accumulate, it can be helpful to automatically label stale issues and PRs and automatically close them after a time.

This workflow labels issues with `stale` if the have not had any activity for 90 days or more, and auto-closes them if they have not had any further activity 60 days after that.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,143 @@
+# Copyright 2025 The Cirq Developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,143 @@
+# Copyright 2025 The Cirq Developers
```
LalehB: WDYT about removing Cirq from this sentence?

LalehB (commented):
Thank you so much, Mike, for adding this. However, I have a concern about this approach:

* What if there is a long-standing issue that needs time to address? It seems to me such issues could be handled manually.

* That said, I think this is a nice idea for stale PRs.

.github/workflows/stale.yml
mhucka: Oh, good grief. Thanks for catching that. Fixed in the next commit.

LalehB:
> > * What if there is a long-standing issue that needs time to address? It seems to me such issues could be handled manually.
> 
> It's true, sometimes issues do take long. There are several options:
> 
> 1. If anything changes (a comment, a label added, etc), then the clock is reset. Often for long-standing issues, _something_ changes, so this often takes care of the problem naturally.
> 2. When an issue is auto-labeled with `stale`, someone can manually remove the label, and that will reset the clock for another 90 days.
> 3. There's a list of labels that makes issues & PRs exempt. Currently the labels are `help wanted` and `health`. We could add others; also, we could add a label specifically to mean "don't ever mark this stale".
> 
> It sounds like a new label might be worth adding. What should it be called? Random ideas off the top of my head: `never stale`, `evergreen`, `undead` (just kidding about that last one), or something else?

I really liked the `undead` but I think something like `never stale` or `do not close` are very clear and explicit which I like 👍🏻

LalehB (approved):
LGTM!

PR #106, opened 2025-08-14
Fix viz

noajshu:
This commit changed the error format:
https://github.com/quantumlib/tesseract-decoder/commit/33bff52db7a7217afaf3845fb75c339180dc88ec

This broke the visualizer. Here we fix the visualizer for compatibility with the new error printing format.

### DIFF
--- a/viz/to_json.py
+++ b/viz/to_json.py
@@ -1,4 +1,18 @@
 # viz/to_json.py
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http:#www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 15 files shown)


viz/to_json.py
```diff
@@ -17,13 +18,20 @@
...
+    return [int(x) for x in INT_RE.findall(text)]
+
 def parse_implicit_list(line, prefix):
     if not line.startswith(prefix):
         raise ValueError(f"Expected line to start with '{prefix}', got: {line}")
     list_part = line[len(prefix):].strip().rstrip(',')
     if not list_part:
         return []
-    return [int(x.strip()) for x in list_part.split(',') if x.strip()]
+    # Be tolerant: accept "1, 2, 3" or "1 2 3"
```
LalehB, line 33: haha "be tolerant" :))

noajshu:
Yes! It would be good to refactor so the visualization code is part of the main source tree and tested properly

PR #108, opened 2025-08-15
Improve the performance of get_detcost() function.

gopinath-vasalamarri:
More information --
https://docs.google.com/spreadsheets/d/17fL_NVWe7XjwGmdBIts75YFuzUmokUyfd_YMEU9ISXY/edit?gid=0#gid=0

### DIFF
--- a/src/tesseract.cc
+++ b/src/tesseract.cc
@@ -17,6 +17,7 @@
 #include <algorithm>
 #include <boost/functional/hash.hpp>  // For boost::hash_range
 #include <cassert>
+#include <cstdint>
 #include <functional>  // For std::hash (though not strictly necessary here, but good practice)
 #include <iostream>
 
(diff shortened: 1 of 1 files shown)


src/tesseract.cc
```diff
@@ -85,24 +86,26 @@ bool Node::operator>(const Node& other) const {
 double TesseractDecoder::get_detcost(
     size_t d, const std::vector<DetectorCostTuple>& detector_cost_tuples) const {
   double min_cost = INF;
+  uint32_t min_det_cost = std::numeric_limits<uint32_t>::infinity();
```
LalehB: Does the `std::numeric_limits<T>::infinity()` applies for `uint32_t`? if it is only for floating-point types then this could become zero and later we might face division by zero issues!

src/tesseract.cc
```diff
@@ -85,24 +86,26 @@ bool Node::operator>(const Node& other) const {
...
     ec = error_costs[ei];
-    if (ec.min_cost >= min_cost) break;
+    if (ec.likelihood_cost * min_det_cost >= min_cost * errors[ei].symptom.detectors.size()) break;
 
     dct = detector_cost_tuples[ei];
     if (!dct.error_blocked) {
-      error_cost = ec.likelihood_cost / dct.detectors_count;
-      if (error_cost < min_cost) {
+      error_cost = ec.likelihood_cost;
+      if (error_cost < min_cost * dct.detectors_count) {
```
LalehB, line 103: I like that you are replacing the div with mul 👏🏻

src/tesseract.cc
```diff
@@ -85,24 +86,26 @@ bool Node::operator>(const Node& other) const {
...
         min_cost = error_cost;
+        min_det_cost = dct.detectors_count;
```
LalehB, line 105: maybe we can have a more discriptive name for this variable, something like `min_cost_det_cnt`?

LalehB (commented):
Thanks so much, @gopinath-vasalamarri 
This optimization is super cool 💯
If you get a chance, could you please also share performance results for larger numbers of shots (10K or 1M) and on other codes (surface code, color code, bivariate bicycle codes)?

src/tesseract.cc
LalehB: replaces it with changed std::numeric_limits<uint32_t>::max()

LalehB (approved):

Issue #92, opened 2025-08-12
Decode to errors seg fault from wrong input type


oscarhiggott:
I accidentally gave decode_to_errors a bitstring of detector outcomes rather than indices of detection events, and got a segmentation fault. Example:

```
from tesseract_decoder import tesseract
import stim


dem = stim.DetectorErrorModel("""
    error(0.1) D0 D1 L0
    error(0.2) D1 D2 L1
    detector(0, 0, 0) D0
    detector(1, 0, 0) D1
    detector(2, 0, 0) D2
""")
config = tesseract.TesseractConfig(dem=dem, det_beam=50)
decoder = config.compile_decoder()
syndrome = [0, 1, 1]
decoder.decode_to_errors(syndrome)
# Segmentation fault (core dumped)
```

Either this should be handled by an exception correctly, or we should switch to have decode_to_errors accept a bitstring of detector outcomes, which would make it consistent with the decode method. WDYT?

LalehB:
@oscarhiggott if it is more intuitive for the researchers I agree to change it (since it is in the python interface)

PR #111, opened 2025-08-20
fixing https://github.com/quantumlib/tesseract-decoder/issues/92 by updating decode_to_errors

noajshu:


src/tesseract.pybind.h
```diff
@@ -243,33 +245,75 @@ void add_tesseract_module(py::module& root) {
...
-           py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>(), R"pbdoc(
+      .def(
+          "decode_to_errors",
+          [](TesseractDecoder& self, const py::array_t<bool>& syndrome) {
+            if ((size_t)syndrome.size() != self.num_detectors) {
+              std::ostringstream msg;
+              msg << "Syndrome array size (" << syndrome.size()
+                  << ") does not match the number of detectors in the decoder ("
+                  << self.num_detectors << ").";
+              throw std::invalid_argument(msg.str());
```
NoureldinYosri: you don't need the stringstream 
```suggestion
             std::string msg = "Syndrome array size (" + std:to_string(syndrome.size())
                  + ") does not match the number of detectors in the decoder ("
                   +  std::to_string(self.num_detectors) + ")."
              throw std::invalid_argument(msg);
```

src/tesseract.pybind.h
```diff
@@ -243,33 +245,75 @@ void add_tesseract_module(py::module& root) {
         config : TesseractConfig
             The configuration object for the decoder.
       )pbdoc")
-      .def("decode_to_errors",
-           py::overload_cast<const std::vector<uint64_t>&>(&TesseractDecoder::decode_to_errors),
-           py::arg("detections"),
-           py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>(), R"pbdoc(
+      .def(
+          "decode_to_errors",
+          [](TesseractDecoder& self, const py::array_t<bool>& syndrome) {
```
NoureldinYosri, line 248: I don't think it is wise to have logic in this part of the code ... having logic here should only happen for simple copying or conversion. lets either modify the C++ decode_to_errors method to have the checks and filter or create a new one that does with tests

src/simplex.pybind.h
```diff
@@ -140,20 +142,43 @@ void add_simplex_module(py::module& root) {
 
         This method must be called before decoding.
       )pbdoc")
-      .def("decode_to_errors", &SimplexDecoder::decode_to_errors, py::arg("detections"),
-           py::call_guard<py::scoped_ostream_redirect, py::scoped_estream_redirect>(), R"pbdoc(
+      .def(
+          "decode_to_errors",
+          [](SimplexDecoder& self, const py::array_t<bool>& syndrome) {
```
NoureldinYosri, line 145: same comments as the other file

AGENTS.md
```diff
@@ -0,0 +1,5 @@
+# Agent Instructions
```
NoureldinYosri, line 1: can the LLM related changes be in a seprate PR?

NoureldinYosri (dismissed):

src/tesseract.pybind.h
noajshu: good idea

noajshu: In a future refactor, let's add methods to the SimplexDecoder and TesseractDecoder APIs which accept a dense representation of the detection events. These will just convert to a sparse representation and call the sparse API methods. The checks for the array size can happen there instead of being duplicated throughout all the pybind glue code for both decoders. WDYT?

Issue #103, opened 2025-08-13
2x perf regression


noajshu:
I just noticed there is like a 2x perf regression on this surface code circuit. I am not sure if I'm doing something differently now than before.
```
git checkout c1b44c8975bc745647ce3d93310a4652e9343507
bazel build src:all && time ./bazel-bin/src/tesseract --sample-num-shots 1000 --circuit testdata/surfacecodes/r\=11\,d\=11\,p\=0.002\,noise\=si1000\,c\=surface_code_X\,q\=241\,gates\=cz.stim  --sample-seed 71734  --threads 64  --beam 20   --num-det-orders 21 --beam-climbing --pqlimit 1000000 --det-order-seed 2384257  --print-stats  --no-revisit-dets
num_shots = 1000 num_low_confidence = 0 num_errors = 0 total_time_seconds = 859.6755540000001

git checkout main
# With --no-det-order-bfs:
bazel build src:all && time ./bazel-bin/src/tesseract --sample-num-shots 1000 --circuit testdata/surfacecodes/r\=11\,d\=11\,p\=0.002\,noise\=si1000\,c\=surface_code_X\,q\=241\,gates\=cz.stim  --sample-seed 71734  --threads 64  --beam 20   --num-det-orders 21 --beam-climbing --pqlimit 1000000 --det-order-seed 2384257  --print-stats  --no-revisit-dets --no-det-order-bfs
num_shots = 1000 num_low_confidence = 0 num_errors = 0 total_time_seconds = 1741.893395


# With --det-order-bfs:
bazel build src:all && time ./bazel-bin/src/tesseract --sample-num-shots 1000 --circuit testdata/surfacecodes/r\=11\,d\=11\,p\=0.002\,noise\=si1000\,c\=surface_code_X\,q\=241\,gates\=cz.stim  --sample-seed 71734  --threads 64  --beam 20   --num-det-orders 21 --beam-climbing --pqlimit 1000000 --det-order-seed 2384257  --print-stats  --no-revisit-dets --det-order-bfs
num_shots = 1000 num_low_confidence = 0 num_errors = 0 total_time_seconds = 1854.489493000001
```

draganaurosgrbic:
Before I leave, I want to leave these commands here to help you:

```
git checkout c1b44c8975bc745647ce3d93310a4652e9343507 (COMMIT FROM THE ISSUE)
[...]

PR #112, opened 2025-08-29
Refactor det order creation, restore optional index-based det ordering

noajshu:
In the paper we used detector indices to sort them and make the det orders. This was actually an accident where we were trying to use the 3d coordinates (xyt) from Stim but used Stim incorrectly (my bad).
This got fixed in a77f5e71a6d3c23981dddb4c5d7fba9a986f3aeb where we started using the actual xyt coordinates.
This seems to offer a better tradeoff between beam and error rate, but also led to issue https://github.com/quantumlib/tesseract-decoder/issues/103 because at the same large beam we get worse performance (likely because there are more distinct orderings which means more ways to get stuck)
We also added BFS det ordering in https://github.com/quantumlib/tesseract-decoder/commit/bf62f847bb3ac8d32054269e013ad8ded30b46e1 and then made it the default in https://github.com/quantumlib/tesseract-decoder/commit/0471d4ea00914c485eb464c6e95c020492ab7ef6
Unfortunately due to a bug in p-ranav's argparse: https://github.com/p-ranav/argparse/issues/413
flags with implicit_value(false) do not work with store_into, which means the `--no-det-order-bfs` arg had no effect. 

Here, we refactor the det order creation to have 3 arguments:
```
  --det-order-bfs                                                         Use BFS-based detector ordering (default if no method specified) 
  --det-order-index                                                       Randomly choose increasing or decreasing detector index order 
  --det-order-coordinate                                                  Random geometric detector orientation ordering 
```

We also eliminate `--no-det-order-bfs` which had no effect anyways.

This is an example where the `--det-order-index` is faster:
```
bazel build src:all
./bazel-bin/src/tesseract --sample-num-shots 100 --circuit testdata/colorcodes/r\=11\,d\=11\,p\=0.001\,noise\=si1000\,c\=superdense_color_code_X\,q\=181\,gates\=cz.stim  --sample-seed 71734  --threads 64  --beam 20 --beam-climbing   --num-det-orders 21  --pqlimit 1000000 --det-order-seed 2384257  --print-stats  --no-revisit-dets --det-order-coordinate 
num_shots = 100 num_low_confidence = 0 num_errors = 0 total_time_seconds = 340.9752209999999

./bazel-bin/src/tesseract --sample-num-shots 100 --circuit testdata/colorcodes/r\=11\,d\=11\,p\=0.001\,noise\=si1000\,c\=superdense_color_code_X\,q\=181\,gates\=cz.stim  --sample-seed 71734  --threads 64  --beam 20 --beam-climbing   --num-det-orders 21  --pqlimit 1000000 --det-order-seed 2384257  --print-stats  --no-revisit-dets --det-order-index
num_shots = 100 num_low_confidence = 0 num_errors = 0 total_time_seconds = 142.994753
```

This is achieved via a new enum class `DetOrder` in utils. It is also now exposed via python API:
```
tesseract_decoder.utils.build_det_orders(
        _DETECTOR_ERROR_MODEL,
        num_det_orders=1,
        method=tesseract_decoder.utils.DetOrder.DetIndex,
        seed=0,

```

Fixes https://github.com/quantumlib/tesseract-decoder/issues/103


oscarhiggott (approved):
Looks great!

PR #113, opened 2025-08-29
refactor of det order helper functions and tests

noajshu:
- added better test for index-based ordering
- refactored the det order function so it uses smaller helpers and a switch statement


noajshu:
Also updated beam climbing behavior to do the max of (beam+1, num_det_orders) # trials

src/utils.cc
```diff
@@ -82,104 +82,124 @@ std::vector<std::vector<size_t>> build_detector_graph(const stim::DetectorErrorM
...
+      if (!visited[start]) {
+        visited[start] = true;
+        q.push(start);
+        perm.push_back(start);
+      }
+      while (!q.empty()) {
+        size_t cur = q.front();
+        q.pop();
+        auto neigh = graph[cur];
+        std::shuffle(neigh.begin(), neigh.end(), rng);
```
LalehB, line 107: I wonder why we are shuffling here? 🤔

src/utils.cc
```diff
@@ -82,104 +82,124 @@ std::vector<std::vector<size_t>> build_detector_graph(const stim::DetectorErrorM
...
+  switch (method) {
+    case DetOrder::DetBFS:
+      return build_det_orders_bfs(dem, num_det_orders, rng);
+    case DetOrder::DetCoordinate:
+      return build_det_orders_coordinate(dem, num_det_orders, rng);
+    case DetOrder::DetIndex:
+      return build_det_orders_index(dem, num_det_orders, rng);
+  }
+  throw std::invalid_argument("Unknown det order method");
+}
```
LalehB, line 201: I wonder if we should have one of them as default 🤔

LalehB (approved):
LGTM!
Thanks Noah, this is pretty cool!

src/utils.cc
noajshu: Ah, this is to generate diversity in the ensemble!

PR #117, opened 2025-08-31
Remove at most two errors per detector

noajshu:


noajshu:
at most two errors is complicating a lot of things with some refactors and I don't think we ever used it in the paper anyways.
I think it will be fairly straightforward to add it back in if we ever need it again :)

PR #132, opened 2025-10-12
Add a Dependabot config file

mhucka:
This adds a `.github/dependabot.yaml` file for configuring [Dependabot](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide) to run on this repo. The configuration makes it check Python and the GitHub Actions used in the CI workflow.

Dependabot regularly looks at the dependencies used in this project (e.g., Python packages) and automatically opens PRs when version updates are available.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -25,7 +25,8 @@ updates:
       - "area/health"
 
   - package-ecosystem: "pip"
-    directory: "/"
+    directories:
+      - "**/*"
     schedule:
       interval: "monthly"
     versioning-strategy: "increase-if-necessary"


.github/dependabot.yaml
```diff
@@ -0,0 +1,35 @@
...
+    # value "/" is given for the directory attribute. Yes, that's confusing.
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    labels:
+      - "area/devops"
+      - "area/health"
+
+  - package-ecosystem: "pip"
+    directory: "/"
```
LalehB: I think we might need to clarify the `"/src/py/"` here? or does the Dependabot searches  recursively deeper in the directory? (`/src/py/` is the location we have the `pip` dependency in the repo)

mhucka: Hmm, yes, you're right, that needs to change. Good catch. The format for recursively searching directories is

```yaml
directories:
  - "**/*"
```

I'll commit an update momentarily.

PR #150, opened 2025-11-08
Remove screenshot from SECURITY.md

mhucka:
It turns out there are different paths to get to the security page on GitHub, and one of them ends up with a broken image. In retrospect, my attempt to be newbie-friendly went overboard with this, and it was a bad idea to try to provide a screenshot because it's too prone to breakage. Let's remove it.


oscarhiggott (approved):
LGTM, though I wonder if just changing the url to a full permalink would fix the issue you have? E.g. `https://raw.githubusercontent.com/quantumlib/tesseract-decoder/7cdc5e158798abb35a44ec99349ee199e8bd5c2b/.github/report-vulnerability-button.png?raw=true`

mhucka:
> LGTM, though I wonder if just changing the url to a full permalink would fix the issue you have? E.g. `https://raw.githubusercontent.com/quantumlib/tesseract-decoder/7cdc5e158798abb35a44ec99349ee199e8bd5c2b/.github/report-vulnerability-button.png?raw=true`

Oh, that probably would work – you're right.

Nonetheless, I'm leaning towards taking it out. If GitHub changes some aspect of that UI, then the screenshot would need to be updated, and the maintenance burden doesn't seem worth it for a page that is probably not going to see a lot of traffic anyway.

PR #179, opened 2026-02-14
Fix `--dem-out` to avoid removing or reformatting errors in the given DEM

noajshu:
The predicted_errors_buffer member in Tesseract and Simplex C++ still stores internal indices (which may be merged and 0-probability errors removed) but now we offer a way to translate back to the original indices in the given DEM. This is used by --dem-out features of tesseract and simplex mains so that we preserve all structure of the given DEM and only update the probabilities. If there are multiple errors that are indistinguishable, we use the count to set the probability of the first occurrence and set the probabilities of all redundant latter errors to be 0.

Also fixes a bug in simplex where we did not compute the cost from the given errors and instead used the predicted errors buffer. And updates simplex mask from and cost from methods to mirror the tesseract style.

The python API transparently translates to the external indices in the original DEM. This is slightly awkward because it disagrees with the C++ api, but hopefully it will not cause problems. In the future it would be great to refactor so that the frequency dem logic is deduped across tesseract and simplex (along with threading and common args). Also curious how to best support dem calibration in a cluster environment where many runs may add up to give one final calibrated DEM.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -117,7 +117,7 @@ double common::merge_weights(double a, double b) {
 }
 
 stim::DetectorErrorModel common::merge_indistinguishable_errors(
-    const stim::DetectorErrorModel& dem, std::vector<size_t>& error_index_mapping) {
+    const stim::DetectorErrorModel& dem, std::vector<size_t>& original_indices) {
   stim::DetectorErrorModel out_dem;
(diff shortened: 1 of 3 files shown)


noajshu:
Going to try this again more carefully. I think the better approach is to always expose in terms of the given DEM indices, for both the C++ and python APIs.

PR #194, opened 2026-02-23
Add a benchmarking script and directory to house test suite and results.

dandragona-dev:
* Builds "main" as a baseline and uses Hyperfine to compare the current directory against this baseline.
* Test suite added with four different error correction code groups. These circuits take a while, so a full run can potentially take hours. Should I just change it to be smaller circuits?
* Creates results directories for each run with whisker plots and stat summaries.

Creates whisker plots like:
results_r18_d18_p0 001_bivariate_bicycle_X


benchmarking/__pycache__/plotting.cpython-313.pyc
noajshu, line 1: Hi @dandragona-dev I don't think these __pycache__ files are intentional, right?

benchmarking/testsuite/r=29,d=29,p=0.001,noise=uniform,c=surface_code,q=3249,gates=cx.stim
noajshu, line 1: It would be nice if we can disclose the type of code etc. that this is, sort of like the file names in testdata/.

benchmarking/__pycache__/plotting.cpython-313.pyc
dandragona: Oh woops, sorry about that! Fixing now.

benchmarking/testsuite/r=29,d=29,p=0.001,noise=uniform,c=surface_code,q=3249,gates=cx.stim
dandragona: Fixed this too.

PR #214, opened 2026-03-03
add demutils as submodule of tesseract

NoureldinYosri:
- turn the files into a python module and export it
- also fixes import paths
- also formats the python files

### DIFF
--- a/src/py/_tesseract_py_util/demutil_test.py
+++ b/src/py/_tesseract_py_util/demutil_test.py
@@ -34,18 +34,6 @@ def test_import_exposes_demutil_submodule():
     assert hasattr(demutil, "decompose_errors")
 
 
-def test_decompose_errors_dispatch_methods():
-    dem = _demo_dem()
-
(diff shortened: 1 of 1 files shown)


src/py/_tesseract_py_util/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    get_component_obs_matching_undecomposed_obs
+from _tesseract_py_util.decompose_errors import \
+    reduce_set_symmetric_difference as reduce_set_symmetric_difference
+from _tesseract_py_util.decompose_errors import \
+    reduce_symmetric_difference as reduce_symmetric_difference
+from _tesseract_py_util.decompose_errors import \
+    undecompose_errors as undecompose_errors
+from _tesseract_py_util.decompose_errors import \
+    undecomposed_error_detectors_and_observables as \
+    undecomposed_error_detectors_and_observables
```
noajshu:
```suggestion
```
Could we avoid exposing the underlying functionality?

PR #215, opened 2026-03-04
Catch invalid symptom in decode_to_errors to avoid segfault

abbrazi:
Put the check in the init loop to avoid a separate loop over the detections.


dandragona (approved):
LGTM, thanks for the guardrail!

Issue #177, opened 2026-02-03
Symbol conflict with highspy causes segfault when both libraries are loaded


plancker:
### Describe the issue

If `tesseract_decoder` is imported before import `highspy` we get a segfault. 
This affects anyone using both `tesseract_decoder` for QEC decoding and `highspy`/`cvxpy` for optimization in the same Python session.

### What version of the software are you using?

0.1.1.dev20260202233842

### How can the issue be reproduced?

```
# Import tesseract_decoder FIRST
from tesseract_decoder import tesseract  # or: import tesseract_decoder

# Then use cvxpy with HiGHS solver
import cvxpy as cp
import numpy as np

x = cp.Variable(3)
A = np.array([[1.0, 2.0, 3.0]])
b = np.array([6.0])
prob = cp.Problem(cp.Minimize(cp.norm1(x)), [A @ x == b])
prob.solve(solver=cp.HIGHS, verbose=False)  # SEGFAULT (exit code 139)
```

**Note:** If the import order is reversed (HiGHS used before importing tesseract_decoder), it works:

```python
# This order works fine
import cvxpy as cp
import numpy as np

x = cp.Variable(3)
prob = cp.Problem(cp.Minimize(cp.norm1(x)), [np.array([[1,2,3]]) @ x == 6])
prob.solve(solver=cp.HIGHS)  # Works
```

from tesseract_decoder import tesseract  # Now safe to import

## Environment

- **tesseract_decoder**: 0.1.1.dev20260202233842 (latest as of Feb 2026)
- **highspy**: 1.12.0
- **cvxpy**: 1.6.0
- **Platform**: macOS 15.3 ARM64 (Apple Silicon)
- **Python**: 3.13

PR #220, opened 2026-03-23
Add a C++ `tesseract_decoder` namespace.

supeterposition:
This prevents namespace collisions when including C++ headers and libraries.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -20,6 +20,8 @@
 #include <string>
 #include <vector>
 
+namespace {
+
 std::string vector_to_string(const std::vector<int>& vec) {
   std::stringstream ss;
   ss << "[";
@@ -33,6 +35,10 @@ std::string vector_to_string(const std::vector<int>& vec) {
   ss << "]";
   return ss.str();
 }
+
+}  // namespace
+
+namespace tesseract_decoder {
 
 std::string common::Symptom::str() const {
   std::string s = "Symptom{detectors=";
@@ -255,3 +261,5 @@ stim::DetectorErrorModel common::dem_from_counts(const stim::DetectorErrorModel&
   }
   return out_dem;
 }
+
+}  // namespace tesseract_decoder
--- a/src/stim_utils.pybind.h
+++ b/src/stim_utils.pybind.h
@@ -7,6 +7,8 @@
 
 #include "stim.h"
 
+namespace tesseract_decoder {
+
 namespace {
 namespace py = pybind11;
 }
@@ -65,5 +67,7 @@ template <typename T>
 void dem_setter(T& config, py::object dem) {
   config.dem = parse_py_object<stim::DetectorErrorModel>(dem);
 }
+
+}  // namespace tesseract_decoder
 
 #endif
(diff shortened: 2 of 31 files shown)


src/stim_utils.pybind.h
```diff
@@ -33,12 +35,12 @@ stim::DemInstructionType parse_dit(std::string dit_str) {
...
 }
 
-stim::DemTarget parse_py_dem_target(py::object py_obj) {
+inline stim::DemTarget parse_py_dem_target(py::object py_obj) {
   return stim::DemTarget::from_text(py::cast<std::string>(py_obj.attr("__str__")()));
 }
 
-stim::DemInstruction parse_py_dem_instruction(py::object py_obj, std::vector<double>& args,
-                                              std::vector<stim::DemTarget>& targets) {
+inline stim::DemInstruction parse_py_dem_instruction(py::object py_obj, std::vector<double>& args,
```
noajshu: what's the reason for the inlining?

noajshu (approved):
Thanks, LGTM!

LalehB (approved):
LGTM!
super nit suggestion: I would recommend a more descriptive name `tesseract_decoder` for the namespace, because google has a tesseract OCR engine that has a namespace tesseract :D I doubt that anyone would use both of them together tho!

src/stim_utils.pybind.h
supeterposition: Actually unrelated to the namespace change, I think it just popped up in my IDE. I can split it out into a separate PR if you'd like. 

Having a function definition in a header file can violate the [one definition rule][odr] if it were to be included from multiple source files and linked together. The function of the `inline` keyword is solely to allow multiple definitions in C++ and get around this issue. So the [Google style guide][cstyle] recommends to always mark function definitions in headers as `inline`(unless they are implicitly inline, for example templated functions).

[cstyle]: https://google.github.io/styleguide/cppguide.html#Defining_Functions_in_Header_Files
[odr]: https://en.wikipedia.org/wiki/One_Definition_Rule

src/common.cc
```diff
@@ -20,6 +20,9 @@
 #include <string>
 #include <vector>
 
+namespace tesseract_decoder {
+namespace {
```
NoureldinYosri, line 23: nit: here and elsewhere, anonymous namespaces don't need to be inside another namespace

src/common.cc
```diff
@@ -241,3 +246,5 @@ stim::DetectorErrorModel common::dem_from_counts(const stim::DetectorErrorModel&
   }
   return out_dem;
 }
+
+}  // namespace tesseract_decoder
```
NoureldinYosri: this file and others need a "\n" at the end

NoureldinYosri (commented):
overall lgtm%nits

src/common.cc
supeterposition: Done.

src/common.cc
supeterposition: Thanks Nour for the review. I moved this instance outside of the project namespace. I reviewed the other anonymous namespace and they reference symbols from `tesseract_decoder` so I'm inclined to keep them nested, but let me know if you feel strongly otherwise.

NoureldinYosri (approved):

noajshu:
This should fix the highspy/cvxpy segfault: https://github.com/quantumlib/tesseract-decoder/issues/177

PR #247, opened 2026-04-17
Reduce number of Dependabot PRs

mhucka:
This changes 2 things:

- Enable grouped updates. Instead of a separate PR for every package,
  Dependabot can combine them into a single "dependency update" PR.

- Limit to direct dependencies. We can use the allow key to restrict it
  to only the packages you explicitly list in the `requirements.in`.


LalehB (approved):
LGTM! Thanks.

PR #254, opened 2026-05-21
Error Sparsification

noajshu:
Adds an optional per-shot error model sparsification heuristic to `tesseract` to get a big speedup with little or no loss in accuracy.
image

This is enabled with `--sparsify-errors`. When this mode is on, Tesseract is allowed to use all errors up to `--sparsify-base-degree k`, then reactivates up to `--sparsify-reactivate-limit M` higher-degree errors up to `--sparsify-max-degree` based on fired-detector overlap. This means we only enable high-degree errors that have a reasonable chance of actually being used. You can also omit `--sparsify-reactivate-limit M` and Tesseract will just choose something that seems to be a reasonable choice for the circuits and `k` values benchmarked below.

An analogy for this would be like if you go to the doctor, they always order the same routine tests (the low-degree errors) but they'll _also_ order a limited number of expensive tests (some high-degree errors) which stand a reasonable chance of illuminating the root cause of symptoms. Here, we infer which high-degree errors might matter from their overlap with the shot's activated detectors.

I benchmarked this on color, surface, and bicycle code circuits with si1000 p=0.001, 0.002 under `tesseract-long-beam` settings. A `--sparsify-base-degree 2` is sufficient for the surface code (it is morally graphlike) while this should be set to `k=3` for color codes and bivariate bicycle codes. I only took data with these values of `k` since the performance at other values seemed much worse in small local runs. I took data at various values of `M` in powers of two. The full results are below:
image
image

It would be annoying for a user to need to tune the `--sparsify-reactivate-limit M` parameter (M = the maximum number of high-degree errors to reactive) to find some Pareto-optimal configuration for each specific circuit. So I attempted to find a reasonable heuristic way of guessing a 'good enough' value of `M` and came up with `M = round((4.5^(k-2) / 3) * num_detectors)`. If we round this to the nearest benchmarked `M`, we can highlight that data point on the tradeoff curve:
image

This is what gives us the quoted change in the first figure above. Note that in some cases the value of M could probably be reduced further, but this heuristic formula provides a simple starting point for further tuning.

For more details of the LER scaling with `M` for each circuit, see the plot below:
image

A few comments for future work:
The implementation of this heuristic could probably be optimized. I'm not sure if the sparsify-reactivate step is a bottleneck, but even during the tesseract search itself the `eneighbors` could be reduced significantly due to reduced connectivity. This optimization is not done here.

It would be interesting if we can discover the best `--sparsify-reactivate-limit k` from inspecting just the DEM.

Full benchmarking dataset: 
[aggregated_results.jsonl.txt](https://github.com/user-attachments/files/28278501/aggregated_results.jsonl.txt)


LalehB (approved):
I love these results! 👏🏻

src/tesseract.cc
```diff
@@ -569,3 +607,69 @@ void TesseractDecoder::decode_shots(std::vector<stim::SparseShot>& shots,
...
+
+  std::sort(candidates.begin(), candidates.end(),
+            [](const OptionalErrorCandidate& a, const OptionalErrorCandidate& b) {
+              if (a.overlap != b.overlap) {
+                return a.overlap > b.overlap;
+              }
+              if (a.degree != b.degree) {
+                return a.degree < b.degree;
+              }
+              if (a.likelihood_cost != b.likelihood_cost) {
```
dandragona, line 656: May not be necessary, but do we need an epsilon value here? since we are comparing doubles.

dandragona (approved):

src/tesseract.cc
noajshu: Good question, since equality of doubles can be unstable even after simple calculations. In this case we are only using the difference to break ties in the sort, so we only need a deterministic result (we are not counting on numerical stability).

dandragona: Ya that's a good point, we just need a tie breaker. Resolving.

PR #256, opened 2026-05-22
Multi pass decoding

oscarhiggott:
## Summary

This adds one- and two-pass decoding with exactly two detector components. Each component has its own Tesseract decoder. In two-pass mode, predictions from the first pass reweight errors in the other component before the final pass. The component decoders are constructed once and reused across shots.

Causal scheduling is the default: only component decodes needed for the final logical prediction are scheduled. Experimental static scheduling runs both components in each pass. More than two components or two passes is out of scope.

## Decoder interface

`MultiPassTesseractDecoder` takes a `MultiPassTesseractConfig`, separately from `TesseractDecoder(TesseractConfig)`. Both, and `SimplexDecoder`, implement the common `Decoder` interface and return a `DecodeResult`. The result carries predictions, cost, low confidence, and an explicit flag indicating whether predicted error indices are populated.

Multipass owns the DEM splitting, component decoding, scheduling, and cross-component reweighting. Generated detector orders resolve separately against each component DEM, using the `DetectorOrder` infrastructure from #277. The CLI chooses the decoder during construction; the worker loop uses the common interface and initialization remains parallel. Execution-plan statistics are computed on demand by `get_execution_plan()`.

## Python and Sinter

The ordinary workflow is:

```python
from multi_pass_sinter_decoders import MultiPassSinterDecoder, get_sinter_decoders

decoder = MultiPassSinterDecoder()
custom_decoders = get_sinter_decoders()
```

Standard Tesseract configuration keywords can be passed directly. Unknown keywords raise. The existing long-beam registry names remain available. Custom X/Z classification is one optional keyword, `detector_basis_classifier`; `detector_classifier` remains a compatibility alias for integer component labels.

GARI, the Python DEM decomposer, and multipass share the detector-basis interface. The named automatic classifier checks, in order:

1. Top-level `measure_basis`.
2. `md.measure_basis`.
3. Top-level `basis`.
4. `md.basis`.
5. Chromobius fourth coordinate: exactly 0/1/2 means X and 3/4/5 means Z.

An invalid reached metadata field blocks fallback. Nonintegral fourth coordinates are rejected. Stim surface-code parity is an explicit named adapter, not an automatic fallback; the decomposer's generic last-coordinate adapter remains available without claiming its labels are necessarily X/Z.

Python resolves classification once and passes a component vector to native code. Users do not need to normalize their DEM manually for Python/Sinter.

## CLI

`--multipass` selects multipass decoding, with `--num-passes 1|2` (default 2) and `--multipass-strategy causal|static`. `--print-multipass-plan` prints model statistics and the schedule on demand. `--multipass --dem-out` is rejected.

The standalone CLI accepts only canonical top-level JSON `measure_basis` tags:

```stim
detector[{"measure_basis":"X"}](0, 0, 0) D0
detector[{"measure_basis":"Z"}](1, 0, 0) D1
```

Canonically tagged circuit `DETECTOR` instructions survive circuit-to-DEM conversion, so tagged `.stim` input works too. Otherwise, `demutil.annotate_detector_bases(...)` produces a canonical DEM in Python. It preserves coordinates, instruction order, repeats, shifts, errors and their tags, and unrelated JSON metadata. Invalid or conflicting existing top-level `measure_basis` values and non-JSON tags that would be overwritten are errors. Lower-priority metadata is retained without requiring agreement. Top-level `measure_basis` is authoritative in both Python and native decoding; `basis`-only DEMs remain supported automatically in Python/Sinter and can be normalized with this helper for the CLI.

## Reweighting and correctness

The reweighting rule is a correlated-matching-style heuristic, not an exact conditional probability. It divides the XOR-combined probability of mechanisms containing both symptoms by the XOR-combined probability of the source symptom. One-sided mechanisms contribute to the denominator; independent coincidences are not included in the numerator. Reweighted probabilities are capped at 0.499 to keep costs positive.

Two-pass decoding requires `merge_errors=True`: the heuristic acts on aggregate symptoms, so applying its probability separately to duplicate unmerged mechanisms would be incorrect. One-pass decoding still supports `merge_errors=False`.

Existing `^` decomposition groups retain their boundaries and tags; each group must belong to one component. For undecomposed mixed-component errors, observable assignment must have exactly one solution. Impossible or ambiguous assignments, logical-only groups, and other detectorless groups are rejected with context rather than silently reinterpreted.

Temporary cost updates validate indices, use deterministic tie ordering, and restore costs even when decoding throws. Repeated and sparse shots must not inherit reweighting state. Reported cost comes only from the final pass; low confidence is aggregated across passes and propagated through Sinter's discard byte. Packed NumPy inputs support arbitrary strides.

## Tests

Coverage includes observable-assignment uniqueness, existing decomposition groups, the reweighting formula and probability cap, merged/unmerged policies, cost restoration and sparse decoding, basis precedence and normalization, tagged circuit input, component detector orders, Sinter discards and strided arrays, and zero-configuration Python/Sinter use. The full `bazel test --jobs=1 src/...` suite passed during implementation, as did the benchmark-workflow and tutorial-sync tests.


src/tesseract_main.cc
```diff
@@ -667,11 +747,14 @@ int main(int argc, char* argv[]) {
         {"pqlimit", args.pqlimit},
         {"num_det_orders", args.num_det_orders},
         {"det_order_seed", args.det_order_seed},
-        {"total_time_seconds", total_time_seconds},
-        {"num_errors", has_obs ? nlohmann::json(num_errors) : nullptr},
-        {"num_low_confidence", num_low_confidence},
+        {"total_time_seconds", global_elapsed},
```
arshpreetmaan: This is writing wall clock time in the JSON instead of the original thread-accumulated time (which is used for downstream per-shot-time calculations and accurate benchmarking)

src/tesseract_main.cc
```diff
@@ -555,35 +578,88 @@ int main(int argc, char* argv[]) {
...
+  auto start_global_time = std::chrono::high_resolution_clock::now();
   size_t shot = parallel_for_shots_in_order(
       shots.size(), args.num_threads,
       [&](size_t thread_index, size_t shot_index) {
-        if (!decoders[thread_index]) {
-          decoders[thread_index] = std::make_unique<TesseractDecoder>(config);
-        }
-        auto& decoder = *decoders[thread_index];
-        auto& error_use = error_use_per_thread[thread_index];
         auto start_time = std::chrono::high_resolution_clock::now();
```
arshpreetmaan: start_time should exclude the decoder construction time (as was done originally) and should be placed just before decoding shots.

src/tesseract.test.cc
arshpreetmaan, line 1: Please restore the existing tests instead of replacing them.

src/tesseract_main.cc
```diff
@@ -555,35 +578,89 @@ int main(int argc, char* argv[]) {
...
+              1e6;
+
+          obs_predicted[shot_index].clear();
+          for (int o : flips) {
+            if (o >= 0 && (size_t)o < num_observables) {
+              obs_predicted[shot_index][o] ^= 1;
+            }
+          }
+          low_confidence[shot_index] = false;
+          cost_predicted[shot_index] = 0;
```
arshpreetmaan: The multipass path hardcodes low_confidence=false and cost=0 because MultiPassTesseractDecoder::decode() returns only observable flips. This hides failures of required final-pass component decoders and reports a misleading zero cost. Could the multipass decoder return a result containing the flips, an aggregate low-confidence status (at least for failed final-pass/observable component) and cost?

src/tesseract_main.cc
arshpreetmaan, line 713: In multipass mode, `error_use_per_thread` is never populated because the decoder returns only observable flips, but `--dem-out` still generates a DEM from these zero counts. Also even if `error_use_per_thread` is returned this would still populate only the mechanisms corresponding to components. Could `--multipass --dem-out` be rejected to be on the safe side?

src/tesseract_main.cc
```diff
@@ -555,35 +578,89 @@ int main(int argc, char* argv[]) {
...
+                       const std::string& tag) -> int {
+    if (tag.find("\"basis\": \"X\"") != std::string::npos) return 0;
+    if (tag.find("\"basis\": \"Z\"") != std::string::npos) return 1;
+    if (coords.size() >= 4) {
+      int c3 = (int)coords[3];
+      if (c3 >= 0 && c3 <= 2) return 0;
+      if (c3 >= 3 && c3 <= 5) return 1;
+    }
+    return 0;
+  };
```
arshpreetmaan: This could hide a partially annotated DEM and produce an incorrect partition by silently assigning the detector to component 0. Could this return -1 and fail validation instead, with an error identifying the unclassified detector?

src/tesseract_main.cc
```diff
@@ -555,35 +578,89 @@ int main(int argc, char* argv[]) {
...
+  tesseract::SchedulingStrategy strategy_val = (args.multipass_strategy == "static")
+                                                   ? tesseract::SchedulingStrategy::Static
+                                                   : tesseract::SchedulingStrategy::Causal;
+
```
arshpreetmaan: Nitpick: Any misspellings such as "causla" or "statci" are silently accepted and defaulted to causal. Could the CLI explicitly accept only "static" and "causal" and report an error for any other value?

src/tesseract_main.cc
aria-googler: done

src/tesseract_main.cc
aria-googler: done

src/tesseract_main.cc
aria-googler: done

src/tesseract_main.cc
aria-googler: done

src/py/tesseract_decoder/utils/__init__.py
arshpreetmaan, line 1: Wait...why is this PR moving the existing src/py/_tesseract_py_util package into src/py/tesseract_decoder/utils? These files appear to be existing utilities with little or no functional change. Could we keep _tesseract_py_util (and src/py/ in whole) and its existing build/test structure unchanged, and add only the new Python files required for multipass?

BUILD
```diff
@@ -19,9 +19,8 @@ py_wheel(
     name="tesseract_decoder_wheel",
     distribution = "tesseract_decoder",
     deps=[
-        "//src:tesseract_decoder",
         "//src/py:generated_stubs",
-        "//src/py/_tesseract_py_util:_tesseract_py_util",
+        "//src/py:tesseract_decoder",
```
arshpreetmaan: Other comment related to _tesseract_py_util structure also affects this

src/multi_pass/multi_pass_tesseract_decoder.cc
```diff
@@ -0,0 +1,407 @@
...
+
+void MultiPassTesseractDecoder::build_causal_schedule() {
+  size_t num_components = component_decoders.size();
+  std::vector<std::set<size_t>> schedule_sets(num_passes);
+
+  // Initial seed: Final pass includes all components that directly affect an
+  // observable.
+  for (size_t i = 0; i < num_components; ++i) {
+    if (component_decoders[i].affects_observable) {
+      schedule_sets[num_passes - 1].insert(i);
```
arshpreetmaan: **Bug:** `--num-passes 0` currently causes a segmentation fault in causal mode and silently performs decoding (and returns a lot of logical errors) in static mode. In causal scheduling, `schedule_sets` has size zero and this line accesses `schedule_sets[num_passes - 1]`. Could the `MultiPassTesseractDecoder` constructor reject `num_passes == 0` before initialization? The CLI and Python/Sinter entry points should also reject values below 1. Please add tests covering zero passes in both static and causal modes.

src/py/tesseract_decoder/utils/__init__.py
aria-googler: sounds good.

src/multi_pass/multi_pass_tesseract_decoder.cc
aria-googler: ok

src/py/tesseract_decoder/utils/__init__.py
arshpreetmaan: resolved with a commit

src/multi_pass/multi_pass_tesseract_decoder.cc
arshpreetmaan: Fixed

src/multi_pass_tesseract_decoder.cc
```diff
@@ -0,0 +1,410 @@
...
+
+  stim::DetectorErrorModel decomposed =
+      decompose_errors_using_generic_classifier(flattened, classifier, true);
+  // std::cout << "DEBUG decomposed:\n" << decomposed << std::endl;
+  stim::DetectorErrorModel merged = merge_indistinguishable_errors(decomposed);
+  // std::cout << "DEBUG merged:\n" << merged << std::endl;
+
+  std::set<int> unique_classes;
+  for (int c : detector_classes)
+    if (c != -1) unique_classes.insert(c);
```
LalehB: The constructor itself never calls `validate_annotations`, and negative classifications are silently excluded here. Those detectors remain in the input syndrome but are assigned to no component, so their detection events are ignored during decoding. The CLI happens to validate first, but the C++ and Python/Sinter APIs do not. Could initialization reject every c < 0, including the detector ID in the error, rather than silently dropping it?

src/multi_pass_tesseract_decoder.h
```diff
@@ -0,0 +1,152 @@
...
+                            SchedulingStrategy strategy = SchedulingStrategy::Static);
+
+  static std::vector<int> classify_detectors(const stim::DetectorErrorModel& dem,
+                                             const DetectorClassifier& classifier);
+
+  MultiPassExecutionPlan get_execution_plan() const;
+  std::vector<int> decode(const std::vector<uint64_t>& detections);
+  MultiPassDecodeResult decode_result(const std::vector<uint64_t>& detections);
+
+  void decode_shots(std::vector<stim::SparseShot>& shots,
```
arshpreetmaan: Is this `decode_shots` method needed in the initial multi-pass API?

The ordinary `TesseractDecoder` already has a convenience method with this name, but this declaration is new for `MultiPassTesseractDecoder` and currently has no definition. The CLI performs its own shot loop, and the Python binding uses `decode_shots_bit_packed`, so this method does not appear to be needed by the current feature.

src/tesseract_main.cc
arshpreetmaan: Could `MultiPassDecodeResult` report only the cost of the components decoded in the final pass?
The current sum also includes first-pass auxiliary predictions.

src/multi_pass_tesseract_decoder.cc
```diff
@@ -0,0 +1,476 @@
...
+  for (size_t pass = 0; pass < num_passes; ++pass) {
+    bool is_final_pass = (pass == num_passes - 1);
+
+    // Decode scheduled components for the current pass layer using persistent
+    // local buffers.
+    for (size_t comp_idx : pass_schedule[pass]) {
+      auto& cd = component_decoders[comp_idx];
+      std::vector<uint64_t> local_dets;
+      for (uint64_t d : detections) {
+        if (cd.component_detectors.count((int)d)) {
```
arshpreetmaan: Could we validate `d < total_global_detectors` before casting it to `int`? `d` is `uint64_t`, so an out-of-range or very large detector is currently narrowed before validation and may otherwise be silently dropped. Please throw an `invalid_argument` identifying the detector.

src/multi_pass_tesseract_decoder.cc
arshpreetmaan: fixed

src/multi_pass_tesseract_decoder.cc
arshpreetmaan: fixed

src/multi_pass_tesseract_decoder.h
arshpreetmaan: fixed

src/tesseract_main.cc
arshpreetmaan: fixed

src/multi_pass_tesseract_decoder.h
```diff
@@ -0,0 +1,149 @@
...
+};
+
+class MultiPassTesseractDecoder {
+ public:
+  MultiPassTesseractDecoder(const stim::DetectorErrorModel& dem, size_t num_passes,
+                            const DetectorClassifier& classifier,
+                            const TesseractConfig& base_config = TesseractConfig(),
+                            size_t num_det_orders = 1, DetOrder det_order_method = DetOrder::DetBFS,
+                            uint64_t seed = 0,
+                            SchedulingStrategy strategy = SchedulingStrategy::Static);
```
arshpreetmaan: The C++ CLI defaults to `causal`, but the C++ constructors and native Python decoder currently default to `static`. Could we use `causal` consistently across all interfaces and expose `strategy` explicitly in the high-level Python constructor? The predefined Sinter decoders already use causal.

arshpreetmaan: fixed

src/tesseract_main.cc
```diff
@@ -667,11 +781,15 @@ int main(int argc, char* argv[]) {
         {"pqlimit", args.pqlimit},
         {"num_det_orders", args.num_det_orders},
         {"det_order_seed", args.det_order_seed},
-        {"total_time_seconds", total_time_seconds},
-        {"num_errors", has_obs ? nlohmann::json(num_errors) : nullptr},
-        {"num_low_confidence", num_low_confidence},
+        {"total_time_seconds", total_time_seconds.load()},
+        {"wall_clock_time_seconds", global_elapsed},
```
arshpreetmaan: Could we remove `wall_clock_time_seconds` and the associated global timer from this PR?

Wall-clock time is not part of the existing decoder JSON convention, and other decoders do not consistently report it. `total_time_seconds` should remain the thread-aggregated decoding time used by the existing tooling. Users can measure wall time externally, and a standardized wall-clock field can be considered separately in a repository-wide PR.

src/multi_pass/multi_pass_sinter_compat.pybind.h
```diff
@@ -0,0 +1,156 @@
...
+
+    for (size_t shot = 0; shot < num_shots; ++shot) {
+      const uint8_t* single_shot_data = detections_data + shot * detections_stride;
+      std::vector<uint64_t> detections;
+      for (uint64_t i = 0; i < num_detectors; ++i) {
+        if ((single_shot_data[i / 8] >> (i % 8)) & 1) detections.push_back(i);
+      }
+
+      std::vector<int> predictions = decoder->decode(detections);
+      uint8_t* single_result_buffer = result_buffer + shot * num_observable_bytes;
```
arshpreetmaan: BUG: This Sinter path calls `decode()` and drops `MultiPassDecodeResult::low_confidence`, so some low-confidence shots can instead be counted as successful. This can make Sinter underestimate the logical failure rate, or equivalently overestimate decoder performance.

Could we open and link a repository-wide follow-up that preserves low-confidence information for both the monolithic (original) and multi-pass Tesseract Sinter bindings?

src/tesseract_main.cc
arshpreetmaan: fixed

src/multi_pass/multi_pass_sinter_compat.pybind.h
arshpreetmaan: Confirmed that this predates PR #256. The existing monolithic `TesseractSinterCompiledDecoder` and legacy `decode_via_files` path also return only observable predictions and drop `low_confidence_flag`.

Tracked in <https://github.com/quantumlib/tesseract-decoder/issues/297>, which covers a shared fix for both the monolithic and multi-pass Sinter integrations. 

Resolving this in favor of the repository-wide issue.

arshpreetmaan (approved):
LGTM! After resolving CI checks related issues and verification of the changes to BUILD files by other reviewers, the PR can be merged.

src/py/multi_pass_sinter_decoders.py
```diff
@@ -0,0 +1,113 @@
...
+        ),
+        "tesseract_multipass_2pass": MultiPassSinterDecoder(
+            num_passes=2,
+            strategy=_core.Causal,
+            det_beam=20,
+            beam_climbing=True,
+            no_revisit_dets=True,
+            merge_errors=True,
+            pqlimit=1000000,
+            num_det_orders=21,
```
oscarhiggott: If these are long-beam settings, maybe "tesseract-long-beam-mono" and similar would be better names.

oscarhiggott:
Should we add these new tests to the CI?

src/tesseract.cc
```diff
@@ -199,6 +199,32 @@ TesseractDecoder::TesseractDecoder(TesseractConfig config_) : config(config_) {
...
+  // Re-sort d2e lists only for affected detectors
+  for (int d : affected_detectors) {
+    if (d >= 0 && (size_t)d < d2e.size()) {
+      std::sort(d2e[d].begin(), d2e[d].end(), [this](size_t idx_a, size_t idx_b) {
+        return error_costs[idx_a].min_cost < error_costs[idx_b].min_cost;
+      });
+    }
+  }
+}
+
```
oscarhiggott: @LalehB @noajshu  it would be helpful if you could review this carefully and check you are happy with it.

src/tesseract_main.cc
```diff
@@ -225,14 +237,15 @@ struct Args {
           std::cout << ")" << std::endl;
         }
       }
-      DetOrder order = DetOrder::DetIndex;
+      DetOrder order = det_order_method;
```
arshpreetmaan: look into this default

src/py/multi_pass_sinter_decoders.py
```diff
@@ -0,0 +1,113 @@
...
+import stim
+import tesseract_decoder as _core
+
+class MultiPassSinterDecoder(sinter.Decoder):
+    """
+    A sinter-compatible Multi-Pass Tesseract Decoder.
+    Wraps the native C++ MultiPassTesseractDecoder.
+    """
+    def __init__(self, num_passes: int = 2, detector_classifier=None,
+                 strategy=_core.Causal, **base_config_kwargs):
```
oscarhiggott: nit (optional): Could improve the docstring to make it clear that all the usual arguments can be passed using `**base_config_kwargs`

src/dem_decomposition.h
```diff
@@ -0,0 +1,84 @@
...
+ * Splits a decomposed DEM into separate DEMs, one for each component ID.
+ */
+std::map<int, stim::DetectorErrorModel> split_dem_by_component(
+    const stim::DetectorErrorModel& dem, const std::function<int(int)>& detector_component_func);
+
+// Returns a detector error model with any error decompositions removed.
+stim::DetectorErrorModel undecompose_errors(const stim::DetectorErrorModel& dem);
+
+// Merges error instructions in a DEM that have the same symptom.
+stim::DetectorErrorModel merge_indistinguishable_errors(const stim::DetectorErrorModel& dem);
```
noajshu: This seems to be redundant with the existing function https://github.com/quantumlib/tesseract-decoder/blob/b20e21f8af3b9346112c88a9d80cc1ac5788015f/src/common.cc#L119 

also this is the only place using `bern_utils`. Ideally we could prune both these new redundant functionalities, WDYT?

oscarhiggott: Yeah I agree. We discussed the duplicated bern_utils earlier but I hadn't noticed that it's only used in another duplicated function! The existing `common::merge_indistinguishable_errors` is better and uses the existing [`merge_weights`](https://github.com/quantumlib/tesseract-decoder/blob/b20e21f8af3b9346112c88a9d80cc1ac5788015f/src/common.cc#L112-L117) which is more numerically stable than the `bernoulli_xor` method also added in this PR in bern_utils (since merge_weights operates directly on the LLRs). Furthermore, I don't think the new `stim::DetectorErrorModel merge_indistinguishable_errors` method introduced in this PR is actually used anywhere! @arshpreetmaan shall we remove the unused `stim::DetectorErrorModel merge_indistinguishable_errors` and also remove the bern_utils files that are no longer needed as a result?

src/multi_pass/error_correlations.cc
```diff
@@ -0,0 +1,151 @@
...
+      for (size_t i = 0; i < components.size(); ++i) {
+        for (size_t j = 0; j < components.size(); ++j) {
+          if (i == j) continue;
+          const auto& hi = components[i];
+          const auto& hj = components[j];
+          if (joint_probs[hi].find(hj) == joint_probs[hi].end()) {
+            joint_probs[hi][hj] = 0.0;
+          }
+          // For small p, joint probability P(A and B) is roughly the sum of p's of bridging errors
+          joint_probs[hi][hj] = joint_probs[hi][hj] * (1 - p) + p * (1 - joint_probs[hi][hj]);
```
noajshu: It seems like this ignores error mechanisms that independently flip only one side, so this is not generally P(hi \land hj).

oscarhiggott: You mean the probability of hi and hj occurring independently? Unless the noise model is very uncorrelated (such that multi-pass decoding is not useful) I wouldn't expect that to impact accuracy much, as the current approach is a good approximation. It's also the same heuristic as used in correlated matching. Perhaps we could just leave a comment documenting the limitation or open an issue for investigating different heurstics?

src/py/multi_pass_sinter_decoders.py
arshpreetmaan: Agreed. Applied the fix.

src/dem_decomposition.h
arshpreetmaan: Yeah I agree. Applied the suggested changes.

src/py/multi_pass_sinter_decoders.py
arshpreetmaan: Fixed

src/multi_pass/multi_pass_tesseract_decoder.h
```diff
@@ -0,0 +1,170 @@
+#ifndef MULTI_PASS_TESSERACT_DECODER_H
```
LalehB, line 15: @noajshu What do you think about moving all the data related to multipass into a subfolder called multi_pass? Or do you prefer this flat structure?

LalehB (approved):

noajshu:
It might be best to create a subfolder:
```

  multi_pass/
    BUILD
    decoder.h
    decoder.cc
    correlation_model.h
    correlation_model.cc
    dem_partition.h
    dem_partition.cc
    detector_classifier.h
    detector_classifier.cc
    sinter_compat.pybind.h

    decoder.test.cc
    correlation_model.test.cc
    dem_partition.test.cc
```

For the future it would be good to consider a broader folder structure reorg for the `src/` folder.

src/multi_pass/multi_pass_tesseract_decoder.h
arshpreetmaan: Applied the suggested changes.

arshpreetmaan:
> It might be best to create a subfolder:
> 
> ```
> 
>   multi_pass/
>     BUILD
>     decoder.h
>     decoder.cc
>     correlation_model.h
>     correlation_model.cc
>     dem_partition.h
>     dem_partition.cc
>     detector_classifier.h
>     detector_classifier.cc
>     sinter_compat.pybind.h
> 
>     decoder.test.cc
>     correlation_model.test.cc
>     dem_partition.test.cc
> ```
> 
> For the future it would be good to consider a broader folder structure reorg for the `src/` folder.

The code is refactored based on these suggestions. Can you PTAL?

src/py/multi_pass_sinter_decoders.py
```diff
@@ -0,0 +1,122 @@
...
+        # 2. Attach the classifier if provided
+        if self.detector_classifier is not None:
+            cpp_decoder.detector_classifier = self.detector_classifier
+        else:
+            def default_classifier(index: int, coords: list[float], tag: str) -> int:
+                import json
+                # Priority 1: Parse JSON tag for "measure_basis" then "basis".
+                # Supports both top-level keys and keys nested under "md".
+                if tag:
+                    try:
```
noajshu: There is a lot of indentation and branching logic here which would be nice to reduce. A helper function like try_load_json which returns None if there is a parse error could be used, for example

src/py/multi_pass_sinter_decoders.py
```diff
@@ -0,0 +1,122 @@
+import sinter
+import stim
+import tesseract_decoder as _core
+
+class MultiPassSinterDecoder(sinter.Decoder):
+    """A Sinter-compatible wrapper around the native multi-pass Tesseract decoder.
+
+    Args:
+        num_passes: Number of passes. Only 1 or 2 are supported.
+        detector_classifier: Optional ``(index, coordinates, tag) -> int`` callable. It must
```
noajshu: Rather than a separate way of specifying and calculating the partitioning of detectors in CSS like error models between GARI, Multipass, decomposer etc., it would be awesome to unify this somehow . gari related code seems to only support chromobius coordinate convention atm [[code](https://github.com/quantumlib/tesseract-decoder/blob/5d2021cc220b755ccf6327e78eb50314b76d82ce/src/py/_tesseract_py_util/gari.py#L261-L296)].

src/py/README.md
```diff
@@ -539,6 +539,60 @@ def get_tesseract_decoder_for_sinter():
...
+`(detector_index, coordinates, tag) -> component_label`. Returning a negative label rejects a
+detector that the callable cannot classify. For example, the built-in coordinate convention can be
+expressed as:
+
+```python
+coordinate_classifier = lambda _index, coordinates, _tag: (
+    0 if len(coordinates) >= 4 and 0 <= coordinates[3] <= 2
+    else 1 if len(coordinates) >= 4 and 3 <= coordinates[3] <= 5
+    else -1
+)
```
noajshu, line 597: is this chromobius convention? Would be good to clarify if so!

src/tesseract_main.cc
```diff
@@ -557,35 +590,136 @@ int main(int argc, char* argv[]) {
...
+    return 1;
+  }
+
+  if (args.multipass_strategy != "static" && args.multipass_strategy != "causal") {
+    std::cerr << "Error: Invalid --multipass-strategy '" << args.multipass_strategy
+              << "'. Expected 'static' or 'causal'." << std::endl;
+    return 1;
+  }
+
+  auto classifier = [](int index, const std::vector<double>& coords,
```
noajshu: I think since we have a python-based classifier for gari, decomposer, multipass, we should unify those under one classifier (in python) for detectors. This is then a separate step and that classification can be used to normalize the DEM to a fixed, modern convention for specifying basis. E.g. with tags or coordinates. This is similar to how the GARI is handled in tesseract CLI. WDYT?
This would be like a preprocessing step where we support those lambda functions, output a circuit / DEM with tags and that is how tesseract CLI understands the detector basis. This 'normalization' helper in python should be sure to avoid unnecessary clobbering of the DEM structure.

src/tesseract_main.cc
```diff
@@ -557,35 +590,136 @@ int main(int argc, char* argv[]) {
...
   std::vector<std::unique_ptr<TesseractDecoder>> decoders(args.num_threads);
+  std::vector<std::unique_ptr<tesseract_decoder::MultiPassTesseractDecoder>> mp_decoders(
+      args.num_threads);
   std::vector<std::vector<size_t>> error_use_per_thread(
       args.num_threads, std::vector<size_t>(original_dem.count_errors()));
   bool has_obs = args.has_observables();
-  size_t num_errors = 0;
-  size_t num_low_confidence = 0;
-  double total_time_seconds = 0;
+  std::atomic<size_t> num_errors(0);
```
noajshu: This is unnecessary and should not be atomic. These counters are incremented by a single main thread only, see below. Although promoting variables to atomic does not make the code malfunction, it gives bad signposting that they are shared variables.

src/tesseract_main.cc
```diff
@@ -557,35 +590,136 @@ int main(int argc, char* argv[]) {
...
+      detector_classes =
+          tesseract_decoder::MultiPassTesseractDecoder::classify_detectors(config.dem, classifier);
+      if (args.print_multipass_plan) {
+        mp_decoders[0] = std::make_unique<tesseract_decoder::MultiPassTesseractDecoder>(
+            config.dem, args.num_passes, detector_classes, config, args.num_det_orders,
+            args.det_order_method, args.det_order_seed, strategy_val, true);
+        std::cerr << mp_decoders[0]->get_execution_plan().str();
+      }
+    } catch (const std::invalid_argument& error) {
+      std::cerr << "Error: " << error.what() << std::endl;
```
noajshu: we prefer to avoid try / catch in C++ code unless absolutely needed, here it is just passing through the error and should be removed. Ditto for anywhere else this is done!

src/tesseract_main.cc
```diff
@@ -557,35 +590,136 @@ int main(int argc, char* argv[]) {
...
       args.num_threads, std::vector<size_t>(original_dem.count_errors()));
   bool has_obs = args.has_observables();
-  size_t num_errors = 0;
-  size_t num_low_confidence = 0;
-  double total_time_seconds = 0;
+  std::atomic<size_t> num_errors(0);
+  std::atomic<size_t> num_low_confidence(0);
+  std::atomic<double> total_time_seconds(0);
+
+  if (args.multipass && !args.dem_out_fname.empty()) {
```
noajshu: this check should be in args validate above.

src/tesseract_main.cc
```diff
@@ -557,35 +590,136 @@ int main(int argc, char* argv[]) {
...
+    std::cerr << "Error: --dem-out is not supported when --multipass is enabled." << std::endl;
+    return 1;
+  }
+
+  if (args.multipass_strategy != "static" && args.multipass_strategy != "causal") {
```
noajshu: ditto -- this logic should be in args validate

src/tesseract_main.cc
```diff
@@ -599,20 +733,20 @@ int main(int argc, char* argv[]) {
         } else if (has_obs && obs_predicted[shot_index] != shots[shot_index].obs_mask) {
           ++num_errors;
         }
-        total_time_seconds += decoding_time_seconds[shot_index];
```
noajshu, line 602: since this is all main thread execution there is no need to use atomic operations

src/tesseract_main.cc
```diff
@@ -627,7 +761,7 @@ int main(int argc, char* argv[]) {
     size_t num_usage_dem_shots = shot;
     if (has_obs) {
       // When we know the obs, we only count non-error shots.
-      num_usage_dem_shots -= num_errors;
+      num_usage_dem_shots -= num_errors.load();
```
noajshu: ditto this is unnecessary atomic stuff

src/tesseract_main.cc
```diff
@@ -669,11 +803,14 @@ int main(int argc, char* argv[]) {
...
-        {"total_time_seconds", total_time_seconds},
-        {"num_errors", has_obs ? nlohmann::json(num_errors) : nullptr},
-        {"num_low_confidence", num_low_confidence},
+        {"total_time_seconds", total_time_seconds.load()},
+        {"num_errors", has_obs ? nlohmann::json(num_errors.load()) : nullptr},
+        {"num_low_confidence", num_low_confidence.load()},
         {"num_shots", shot},
         {"num_threads", args.num_threads},
+        {"multipass", args.multipass},
+        {"strategy", args.multipass_strategy},
```
noajshu: prefer multipass_strategy and multipass_num_passes

src/tesseract_main.cc
```diff
@@ -690,11 +827,11 @@ int main(int argc, char* argv[]) {
   }
   if (print_final_stats) {
     std::cout << "num_shots = " << shot;
-    std::cout << " num_low_confidence = " << num_low_confidence;
+    std::cout << " num_low_confidence = " << num_low_confidence.load();
```
noajshu: ditto, unnecessary atomic stuff

noajshu (commented):
Left a few comments. One issue you will have with the C++ CLI is it is used to only dealing with TesseractDecoder instances. But now the decoder could be a multipass class instance. There are a few good possible polymorphism patterns worth looking into e.g. a parent / child or sibling class relationship or templating etc.

noajshu (approved):
LGTM
@oscarhiggott will this break your workflow?

PR #257, opened 2026-05-27
Expose error model sparsification to python module, update sinter decoders dict

noajshu:
Exposes per-shot error model sparsification to the Python module, CLI, and Sinter compatibility layer.

Changes:
- Added Python sparsification config fields: `sparsify_errors`, `sparsify_base_degree`, `sparsify_max_degree`, and `sparsify_reactivate_limit`.
- Added module-level `tesseract.suggest_sparsify_reactivate_limit(num_detectors, sparsify_base_degree)` for documenting and testing the built-in heuristic.
- Resolves `sparsify_reactivate_limit == -1` only when preparing the decoder, and reports the effective resolved value in CLI stats output.
- Caps the auto reactivation limit to the compiled error count and avoids overflow in the heuristic computation.
- Added validation for sparsification parameters: base degree must be positive when sparsification is enabled, max degree must not contradict base degree, and reactivation limit must be `>= -1`.
- Kept CLI sparsification semantics explicit: user-provided nonnegative `--sparsify-reactivate-limit` is preserved, while `-1` or an omitted value uses auto resolution at decoder initialization.
- Changed the C++ CLI, Python config, and Sinter decoder defaults to DetIndex ordering. `--det-order-bfs` remains available for the old BFS behavior.
- Extended `TesseractSinterDecoder` to support sparsification config while preserving pickle/serialization compatibility for existing distributed Sinter runs.
- Registered readable sparsified decoders in the Sinter decoders dictionary: `tesseract-long-beam-sparsify-color-code-like`, `tesseract-long-beam-sparsify-surface-code-like`, `tesseract-short-beam-sparsify-color-code-like`, and `tesseract-short-beam-sparsify-surface-code-like`.
- Added README guidance for choosing non-sparsified, surface-code-like sparsified, color-code-like sparsified, long-beam, and short-beam variants.
- Added Python and C++ unit tests for sparsify config, validation, heuristic suggestion, Sinter propagation, auto-limit capping, and end-to-end decoding.


src/py/README.md
```diff
@@ -59,6 +66,21 @@ print(f"Custom configuration no-revisit detection events: {config2.det_beam}")
...
+    dem=dem,
+    det_beam=20,
+    beam_climbing=True,
+    sparsify_errors=True,
+    sparsify_base_degree=3,
+    sparsify_reactivate_limit=-1,
+)
+decoder = config3.compile_decoder()
+print(
+    "Resolved sparsify reactivation limit:",
```
mhucka, line 81: Not 100% sure, but I suspect there should be a space after the colon.

```suggestion
    "Resolved sparsify reactivation limit: ",
```

src/py/tesseract_test.py
```diff
@@ -286,5 +286,139 @@ def test_test_simplex_decode_batch_with_mismatched_syndrome_size():
...
+
+
+@pytest.mark.parametrize(
+    "kwargs, message",
+    [
+        (
+            {"sparsify_reactivate_limit": -2},
+            "sparsify_reactivate_limit must be >= -1",
+        ),
+        ({"sparsify_max_degree": -2}, "sparsify_max_degree must be >= -1"),
```
mhucka: Not a big deal, just a nit: the indentation seems inconsistent here.

src/tesseract.cc
```diff
@@ -71,6 +72,15 @@ std::string TesseractConfig::str() {
   return ss.str();
 }
 
+int suggest_sparsify_reactivate_limit(size_t num_detectors, int sparsify_base_degree) {
+  if (sparsify_base_degree < 0) {
+    throw std::invalid_argument("sparsify_base_degree must be >= 0.");
+  }
+  double k = sparsify_base_degree;
+  return static_cast<int>(
+      std::round((std::pow(4.5, k - 2.0) / 3.0) * static_cast<double>(num_detectors)));
```
mhucka: Is there a risk of overflow in the `std::pow` expression, if an unexpectedly large `k` is passed in? (Even if by accident.)

src/tesseract_main.cc
```diff
@@ -225,8 +225,10 @@ struct Args {
           std::cout << ")" << std::endl;
         }
       }
-      DetOrder order = DetOrder::DetBFS;
-      if (det_order_index) {
+      DetOrder order = DetOrder::DetIndex;
```
mhucka, line 228: This looks like it's using a default detector ordering of `DetOrder::DetIndex`, but in src/tesseract_sinter_compat.pybind.h line 137, it looks like it may be using a default of `DetOrder::DetBFS`. Similarly, in src/tesseract.pybind.h it looks like `DetOrder::DetBFS` may be the default. Does this inconsistency matter?

mhucka (commented):
I'm not qualified to evaluate the algorithmic or theoretical correctness, but I tried to look at this from the a basic code perspective. I had a couple of minor items, noted in-line.

In addition, one thing that I don't see (and this may just be due to ignorance of Tesseract Decoder) is an update to the `.pyi` files. This PR does seem to introduce new properties on some objects, which suggests the .pyi files may become out of date. Not 100% sure – just flagging it for your attention.

src/py/README.md
```diff
@@ -488,6 +510,17 @@ The Tesseract Python interface is compatible with the Sinter framework, which is
 
 #### The TesseractSinterDecoder Object
 All Sinter examples rely on this utility function to provide the Sinter-compatible Tesseract decoder.
+The default decoder dictionary also includes sparsified variants:
+`tesseract-long-beam-sparsify3`, `tesseract-long-beam-sparsify2`,
+`tesseract-short-beam-sparsify3`, and `tesseract-short-beam-sparsify2`.
+
+As a quick rule of thumb, use the non-sparsified decoders as the safest baseline. Use `sparsify2`
+for surface-code-like or mostly graphlike DEMs, and use `sparsify3` for color-code,
```
LalehB: super nit: wdyt about using a name like `sparsify-surface-code-like` instead of `sparcify2`? I know it becomes longer which is not the best but something more descriptive might be good.

src/py/README.md
noajshu: Thanks, I think python adds a space due to the print(foo, bar) syntax

src/tesseract.cc
noajshu: Good catch, let's handle this more carefully!

src/tesseract_main.cc
noajshu: Aha, so now I switched it to be detindex everywhere, thanks for the catch!

LalehB (approved):
LGTM!

src/py/README.md
mhucka: > Thanks, I think python adds a space due to the print(foo, bar) syntax

Oh, ack, you're right; I missed the comma at the end of the string. NVM!

mhucka (approved):

src/py/README.md
noajshu: This is a good idea. Let me think and ask some others about the naming as well.

noajshu:
Thanks Mike and Laleh for the reviews!

src/py/README.md
noajshu: Good idea! Went with surface-code-like and color-code-like

Issue #258, opened 2026-05-27
Run error: libc++abi: terminating due to uncaught exception of type std::length_error: vector


LaraTatli18:
### Describe the issue

Bazel build succeeded but issues running Tesseract
Error when ./bazel-bin/src/tesseract --circuit testdata/surfacecodes/r=5,d=5,p=0.0001,noise=si1000,c=surface_code_
Z,q=49,gates=cz.stim --sample-num-shots 1000 --out predictions.01 --out-format 01

Same issue with d5r5colorcode_p001.stim

Exact output: libc++abi: terminating due to uncaught exception of type std::length_error: vector
zsh: abort

LalehB:
Hi @LaraTatli18 , can you please check and see if you still see this issue? (I addressed it in #262 )

LaraTatli18:
> Hi [@LaraTatli18](https://github.com/LaraTatli18) , can you please check and see if you still see this issue? (I addressed it in [#262](https://github.com/quantumlib/tesseract-decoder/pull/262) )

Hi @LalehB, sorry I missed this, yes it seems to be building and decoding correctly for the test case above! Will also confirm with larger cases

PR #262, opened 2026-06-15
fix issue #258: guard error_chain_arena.reserve() against SIZE_MAX pqlimit

LalehB:
Fixing issue #258 
When `--pqlimit` is not specified, it defaults to `std::numeric_limits<size_t>::max()`
(SIZE_MAX). Calling `vector::reserve(SIZE_MAX)` exceeds the vector's max_size()
and throws std::length_error, crashing the process.
Fix by only calling reserve() when pqlimit is a finite (non-SIZE_MAX) value.


noajshu (approved):
Thank you, this is indeed an important edge case!

PR #263, opened 2026-06-18
Add sparsify error benchmarking artifacts

noajshu:
Adds the benchmarking artifacts used for the error sparsification work in #254 under `benchmarking/sparsify_errors/`.

This includes:
- the Slurm submission script used for the sparsify-errors sweep
- the aggregated JSONL benchmark dataset
- the plotting/analysis script used to generate the PR figures
- the generated PDF plots


arshpreetmaan (approved):
All good.

benchmarking/sparsify_errors/plot.py
```diff
@@ -0,0 +1,804 @@
...
+            return v
+            
+    actual_path = circuit_path
+    if not os.path.exists(actual_path):
+        alt_path = os.path.join('testdata', os.path.basename(circuit_path))
+        if os.path.exists(alt_path):
+            actual_path = alt_path
+        else:
+            _E_CACHE[circuit_path] = (1, 1)
+            return (1, 1)
```
LalehB: when we are returning (1, 1) here when the circuit file is missing is it possible that it would generate plausible-looking but incorrect plots for some cases? wdyt of raising an error by default, or adding an explicit `--allow-missing-circuits` mode for offline reproduction?

benchmarking/sparsify_errors/plot.py
```diff
@@ -0,0 +1,804 @@
...
+import matplotlib.lines as mlines
+import matplotlib.cm as cm
+import numpy as np
+from scipy.stats import binomtest
+import stim
+
+_E_CACHE = {}
+
+# Fallback mapping extracted directly from your provided output.
+_FALLBACK_DETECT_MAP = {
```
LalehB: For this fallback map that depends on exact circuit filenames. wdyt if we either compute these metrics from the circuit files directly or store E and num_detectors in aggregated_results.jsonl? That would make the analysis less coupled to one benchmark.

LalehB (approved):

benchmarking/sparsify_errors/plot.py
noajshu: Good catch—I’ll remove the silent fallback and make missing or invalid metadata a hard error.

benchmarking/sparsify_errors/plot.py
noajshu: Agreed—I’ll store the required metrics in the JSONL and remove the filename-based map.

PR #268, opened 2026-07-06
Add CPU trellis decoder

noajshu:
## Summary
Adds a `tesseract_trellis` decoder binary and library.
I built this on a branch and opened a PR earlier (https://github.com/quantumlib/tesseract-decoder/pull/230) but closed it because it grew too big and ocmplicted.

But I do need this tesseract trellis binary to live somewhere and anywhere now we should land it.
In the meantime at QEC26 I saw Leverrier's talk on the [frontier decoder](https://arxiv.org/pdf/2606.20513) which is quite similar conceptually. There are some differences but anyways it is a nice idea and I look forward to comparing to the different future cost estimates in their paper.

Usage example:

```bash
bazel build src:tesseract_trellis

./bazel-bin/src/tesseract_trellis \
  --circuit path/to/circuit.stim \
  --sample-num-shots 1000 \
  --sample-seed 1234 \
  --threads 1 \
  --ranking-mode future-detcost \
  --beam 1024 \
  --obs-probs-out obs_probs.bin
```

I highly recommend to use future-detcost ranking mode to allow good performance at the smallest beam sizes.


src/tesseract_trellis_main.cc
```diff
@@ -0,0 +1,506 @@
...
+    std::cerr << err.what() << std::endl;
+    std::cerr << program;
+    return EXIT_FAILURE;
+  }
+
+  args.validate();
+  TesseractTrellisConfig config;
+  std::vector<stim::SparseShot> shots;
+  std::unique_ptr<stim::MeasureRecordWriter> writer;
+  args.extract(config, shots, writer);
```
mhucka, line 338: It looks like args.validate() and extract() can throw error, but the calls here are not wrapped in a try-catch. Should they be? I'm not a C++ person but IIRC, uncaught exceptions will result in a core dump. If so, it would be more user-friendly to print an error and exit with EXIT_FAILURE.

src/tesseract_trellis_cli_test.sh
```diff
@@ -0,0 +1,98 @@
...
+"${binary}" \
+  --dem "${one_obs_dem}" \
+  --in "${shots_01}" \
+  --in-format 01 \
+  --dem-out "${work_dir}/unsupported.dem" \
+  >"${work_dir}/dem_out.stdout" \
+  2>"${work_dir}/dem_out.stderr"
+dem_out_status=$?
+set -e
+[[ "${dem_out_status}" -ne 0 ]]
```
mhucka: If the binary terminates on an uncaught exception, will this silently pass? (Will the return be non-0 in that case?)

Could this test be made more robust?

src/tesseract_trellis_main.cc
```diff
@@ -0,0 +1,506 @@
...
+        return !has_obs || num_errors < args.max_errors;
+      });
+
+  if (!args.obs_probs_out_fname.empty()) {
+    std::ofstream out(args.obs_probs_out_fname, std::ios::binary);
+    if (!out.is_open()) {
+      throw std::invalid_argument("Failed to open " + args.obs_probs_out_fname);
+    }
+    for (size_t k = 0; k < shot; ++k) {
+      out.write(reinterpret_cast<const char*>(&obs_probability_predicted[k]), sizeof(double));
```
mhucka: Is there a reason why this is not written out using a single operation? It doesn't look like any function is being called to generate results in the loop, which would mean the array of probabilities is unchanging in this loop, which suggests the whole array could be written in one go. As it is, the loop will make `k` number of system calls, possibly unnecessarily.

src/tesseract_trellis_main.cc
```diff
@@ -0,0 +1,506 @@
...
+          throw std::invalid_argument("Could not open the file: " + out_fname);
+        }
+      }
+      writer = stim::MeasureRecordWriter::make(predictions_file, predictions_out_format.id);
+      writer->begin_result_type('L');
+    }
+  }
+};
+
+int main(int argc, char* argv[]) {
```
mhucka, line 259: This main is very long. Could it be broken up into small pieces, for better modularity and readability?

src/tesseract_trellis_main.cc
```diff
@@ -0,0 +1,506 @@
...
+    if (has_obs) {
+      stats_json["num_errors"] = num_errors;
+    } else {
+      stats_json["num_errors"] = nullptr;
+    }
+    if (args.stats_out_fname == "-") {
+      std::cout << stats_json << std::endl;
+      print_final_stats = false;
+    } else {
+      std::ofstream out(args.stats_out_fname, std::ofstream::out);
```
mhucka, line 478: This call is not checked for errors, such as nonexistent paths. Is there a chance that it will will silently fail and no output will be written?

src/tesseract_trellis.cc
```diff
@@ -0,0 +1,1283 @@
...
+  ~FinalizeKeptStateStatsOnExit();
+};
+
+std::vector<Fault> parse_faults(const std::vector<common::Error>& errors, size_t num_observables) {
+  std::vector<Fault> faults;
+  faults.reserve(errors.size());
+  for (size_t error_index = 0; error_index < errors.size(); ++error_index) {
+    const auto& error = errors[error_index];
+    const double p = error.get_probability();
+    if (p <= 0) {
```
mhucka, line 124: Should there be an explicit check for `p` being NaN or other invalid values here? (`NaN` <= 0` will evaluate to false, so the test would succeed, but then the log functions below would be called with the value.)

src/tesseract_trellis_main.cc
```diff
@@ -0,0 +1,506 @@
...
+      }
+      writer = stim::MeasureRecordWriter::make(predictions_file, predictions_out_format.id);
+      writer->begin_result_type('L');
+    }
+  }
+};
+
+int main(int argc, char* argv[]) {
+  std::cout.precision(16);
+  argparse::ArgumentParser program("tesseract_trellis");
```
mhucka, line 261: I would suggest also adding a `--help` option (unless the Args library provides one by default).

src/tesseract_trellis_main.cc
```diff
@@ -0,0 +1,506 @@
...
+      }
+      std::vector<stim::SparseShot> shots_in_range(shots.begin() + shot_range_begin,
+                                                   shots.begin() + shot_range_end);
+      std::swap(shots_in_range, shots);
+    }
+
+    if (!out_fname.empty()) {
+      stim::FileFormatData predictions_out_format = stim::format_name_to_enum_map().at(out_format);
+      FILE* predictions_file = stdout;
+      if (out_fname != "-") {
```
mhucka, line 247: What happens if `out_name` is `-`?

mhucka (changes requested):
This is not as throughout a review as it should be; this is a long PR and I'm kind of out of my depth on the technical and C++ details. Still, I did what I could. Hopefully the comments are more helpful than annoying …

src/tesseract_trellis_main.cc
noajshu: Thanks Mike! if you try it out you should see it works in a way that is pretty user-friendly. E.g.:
```
(tesseract) noah@Noahs-Laptop tesseract-decoder % ./bazel-bin/src/tesseract
libc++abi: terminating due to uncaught exception of type std::invalid_argument: Must provide at least one of --circuit or --dem
zsh: abort      ./bazel-bin/src/tesseract
```

src/tesseract_trellis_main.cc
noajshu: switched it, although note that ofstream is buffered

src/tesseract_trellis_main.cc
noajshu: no, but also this is consistent with the other two binaries (tesseract and simplex)

src/tesseract_trellis_main.cc
noajshu: I removed some of the logging to make it a bit shorter.

noajshu:
> This is not as throughout a review as it should be; this is a long PR and I'm kind of out of my depth on the technical and C++ details. Still, I did what I could. Hopefully the comments are more helpful than annoying …

Thank you @mhucka ! This is very helpful. I addressed many of the comments.

src/tesseract_trellis.cc
```diff
@@ -0,0 +1,1288 @@
...
+
+#include "tesseract_trellis.h"
+
+#include <algorithm>
+#include <array>
+#include <bit>
+#include <chrono>
+#include <cmath>
+#include <cstdint>
+#if defined(__BMI2__) && \
```
mhucka, line 23: This is a compile-time flag. If the wheels are built on a machine with BMI2 enabled, but a user runs it on a machine without BMI2, will it cause problems?

mhucka (approved):
One last detail. Optional. Otherwise LGTM!

src/tesseract_trellis.cc
noajshu: Yes, that's possible. If we expose this to the python module then we may run into things like this eventually. But for now as a C++ prototype it should be OK.

noajshu:
Thanks again for the review @mhucka !

PR #271, opened 2026-07-22
fix: remove 64-bit observable scaling limit in CLI evaluation path

arshpreetmaan:
**Motivation:**
The core Tesseract C++ engine and its Python bindings handle $>64$ observables dynamically and correctly, which is critical for evaluating high rate QEC circuits incorporating hundreds of logical operators. However, the standalone C++ CLI executables (`tesseract_main`, `simplex_main`, etc.) artificially introduced a mathematical ceiling. The CLI previously tracked predicted observable states using a single primitive `uint64_t` bitmask to optimize evaluations and raw binary output writing. When scaling past 64 observables natively, this primitive silently overran its bounds, corrupting the binary output and causing the CLI's internal false-negative trackers to incorrectly evaluate structural failures (e.g., reporting `num_errors = num_shots`).

This PR strips away the artificial CLI ceiling, restoring parity between the C++ binary tools and the Python bindings. 

**Key Changes:**
1. **Dynamic Bitmask Scaling:** Replaced all hard-coded `uint64_t` primitive tracking logic within the CLI wrappers (`tesseract_main.cc`, `simplex_main.cc`, and `tesseract.perf.cc`) with STIM's natively scalable `stim::simd_bits<64>` tracking structures.
2. **Memory Safe Hot-Paths:** Migrated the CLI's fast-path evaluations and prediction updates to use in-place `.clear()` techniques on the pre-allocated `simd_bits` vectors, actively avoiding exponential heap re-allocations during hot loop iteration.
3. **Structured Binary Exports:** Fixed a memory overrun bug in the `.b8` binary writer. The `writer->write_bits` logic now rigorously calls `.u8` to extract the raw allocated byte pointer from the `simd_bits` class, ensuring cleanly formatted Sinter-compatible binary dumps unbounded by length.
4. **Integration Testing:** Added an explicit test case to check scaling capabilities above the 64-observable limit.

### DIFF
--- a/src/simplex_main.cc
+++ b/src/simplex_main.cc
@@ -479,6 +479,8 @@ int main(int argc, char* argv[]) {
           std::cout << "cost = " << cost_predicted[shot_index] << std::endl;
           std::cout.flush();
         }
+        // Disable early termination due to \`--max-errors\` when we don't have the ground-truth
+        // observables
         return !has_obs || num_errors < args.max_errors;
       });
 
--- a/src/simplex_main.cc
+++ b/src/simplex_main.cc
@@ -422,7 +422,9 @@ int main(int argc, char* argv[]) {
   std::vector<stim::SparseShot> shots;
   std::unique_ptr<stim::MeasureRecordWriter> writer;
   args.extract(config, shots, writer);
-  std::vector<uint64_t> obs_predicted(shots.size());
+  size_t num_observables = config.dem.count_observables();
+  std::vector<stim::simd_bits<64>> obs_predicted(shots.size(),
+                                                 stim::simd_bits<64>(num_observables));
   std::vector<double> cost_predicted(shots.size());
   std::vector<double> decoding_time_seconds(shots.size());
   const stim::DetectorErrorModel original_dem = config.dem.flattened();
@@ -432,7 +434,6 @@ int main(int argc, char* argv[]) {
   bool has_obs = args.has_observables();
   size_t num_errors = 0;
   double total_time_seconds = 0;
-  size_t num_observables = config.dem.count_observables();
   size_t shot = parallel_for_shots_in_order(
       shots.size(), args.num_threads,
       [&](size_t thread_index, size_t shot_index) {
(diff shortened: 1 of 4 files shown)


noajshu:
Thanks!

I noticed that, even before this PR, the handling of num_errors when observable ground truth is absent is nonsensical.

This PR makes the behavior worse: when has_obs == false, shots[shot_index].obs_mask is zero-sized, while obs_predicted is allocated to the DEM’s observable count. Since simd_bits equality compares sizes, the masks are always unequal whenever the DEM has observables. Tesseract therefore counts every non-low-confidence shot as an error, and Simplex counts every shot as an error.

In this mode, the binaries also print num_errors under --print-stats, write it to --stats-out, and use it for --max-errors. The ordinary final summary already correctly omits it.

Could you make all comparison, reporting, and --max-errors logic conditional on has_obs? --stats-out should probably omit num_errors or use null when no observable truth is available.

arshpreetmaan:
Thanks for catching that @noajshu ! I have applied the proposed fix.

src/simplex_main.cc
```diff
@@ -447,31 +448,38 @@ int main(int argc, char* argv[]) {
...
         total_time_seconds += decoding_time_seconds[shot_index];
         if (args.print_stats) {
-          std::cout << "num_shots = " << (shot_index + 1) << " num_errors = " << num_errors
-                    << " total_time_seconds = " << total_time_seconds << std::endl;
+          std::cout << "num_shots = " << (shot_index + 1);
+          if (has_obs) {
+            std::cout << " num_errors = " << num_errors;
+          } else {
+            std::cout << " num_errors = N/A";
+          }
```
noajshu, line 477: formatting nit
```suggestion
          }
```

src/simplex_main.cc
```diff
@@ -447,31 +448,38 @@ int main(int argc, char* argv[]) {
...
+          std::cout << " total_time_seconds = " << total_time_seconds << std::endl;
           std::cout << "cost = " << cost_predicted[shot_index] << std::endl;
           std::cout.flush();
         }
-        return num_errors < args.max_errors;
+        return !has_obs || num_errors < args.max_errors;
```
noajshu, line 484:
```suggestion
        // Disable early termination due to `--max-errors` when we don't have the ground-truth observables
        return !has_obs || num_errors < args.max_errors;
```

src/tesseract_main.cc
```diff
@@ -574,36 +575,42 @@ int main(int argc, char* argv[]) {
...
           std::cout << "num_shots = " << (shot_index + 1)
-                    << " num_low_confidence = " << num_low_confidence
-                    << " num_errors = " << num_errors
-                    << " total_time_seconds = " << total_time_seconds << std::endl;
+                    << " num_low_confidence = " << num_low_confidence;
+          if (has_obs) {
+            std::cout << " num_errors = " << num_errors;
+          } else {
+            std::cout << " num_errors = N/A";
+          }
```
noajshu, line 608: ditto here:
```suggestion
          }
```

src/tesseract_main.cc
```diff
@@ -574,36 +575,42 @@ int main(int argc, char* argv[]) {
...
+          std::cout << " total_time_seconds = " << total_time_seconds << std::endl;
           std::cout << "cost = " << cost_predicted[shot_index] << std::endl;
           std::cout.flush();
         }
-        return num_errors < args.max_errors;
+        return !has_obs || num_errors < args.max_errors;
```
noajshu, line 615:
```suggestion
        // Disable early termination due to `--max-errors` when we don't have the ground-truth observables
        return !has_obs || num_errors < args.max_errors;
```

noajshu (approved):
LGTM with a few small comments!

arshpreetmaan:
I addressed the comments. Thank you for reviewing @noajshu !

LalehB (approved):
LGTM!
thanks Arshpreet.

PR #287, opened 2026-08-06
Drop support for legacy systems to fix CI

NoureldinYosri:
- Drop py3.10 since it is no longer supported by scientific libraries and break the CI
- normalize the yaml file to avoid warnings
- drop support for legacy system since we can't build them using google's recommended actions versions
- apply zizmor


noajshu:
Thanks, LGTM!  I realize some of our CI checks currently only run when we land, so we'll find out soon if there are any more issues.

noajshu (approved):

PR #289, opened 2026-08-07
Remove citation.cff updater workflow

mhucka:
The Google Zizmor security scanner has issues with it, and since it's not clear it's doing all that much good anyway, let's just remove the workflow for now.


viathor:
What are the security issues raised by the scanner?

mhucka:
> What are the security issues raised by the scanner?

@viathor It seemed to think that some of the variable expansions could be exploited:

image

This came about from allowing user input in the manual invocation trigger (workflow_dispatch). Now, this workflow _could_ be rewritten to remove that part; when I wrote it originally, I thought it might be useful to have that capability, but no one used it in any of the other repositories where I had put this workflow, so it seems to be more trouble than it's worth. 

Plus, I'm not sure if this workflow will update CITATION.cff at the right point in a release. I had removed it from all other repos because of that. I was surprised it was still in the Tesseract repo, which I assumed was my fault, which is why I thought it wouldn't be missed if I just deleted it.

PR #295, opened 2026-08-10
Parametrize Bazel `disk-cache` by `matrix.os`

mhucka:
The Bazel disk cache should probably be scoped by the `matrix.os` value and the Python version in matrix jobs. This observation was spurred by https://github.com/quantumlib/tesseract-decoder/pull/277#issuecomment-5229263559, although it is not yet clear whether narrowing the scope will in fact fix the failure in that PR.


LalehB (approved):
Thank you Mike!

PR #298, opened 2026-08-11
Fix portable wheel builds and validate wheels in PR CI

### LINKED ISSUE #291: pypi advertises incompatible version to out-of-date linux
### Describe the issue

reproduction:
```
docker run --rm --platform linux/amd64 quay.io/pypa/manylinux2014_x86_64 bash -lc '
set -euxo pipefail

PY=/opt/python/cp311-cp311/bin/python

echo "Test system:"
ldd --version | head -n 1

"$PY" -m pip install --upgrade pip
"$PY" -m pip install \
    --pre \
    --no-cache-dir \
    --only-binary=:all: \
    tesseract-decoder

"$PY" -c "
import tesseract_decoder
print(\"PASS: tesseract_decoder imported successfully\")
"
'
```
output:
```
Traceback (most recent call last):
  File "<string>", line 2, in <module>
ImportError: /lib64/libc.so.6: version `GLIBC_2.38' not found (required by /opt/python/cp311-cp311/lib/python3.11/site-packages/tesseract_decoder.so)
```

### What version of the software are you using?

_No response_

### How can the issue be reproduced?

_No response_

### LINKED ISSUE #296: CI does not test wheel build process
### Describe the issue

Although we should not publish wheels upon a PR being opened, we should likely test that the code to build the wheel is not broken. For example, this PR had green checks:
https://github.com/quantumlib/tesseract-decoder/pull/293

but when landed it failed:
https://github.com/quantumlib/tesseract-decoder/actions/runs/31423476383/job/93569588146

### What version of the software are you using?

_No response_

### How can the issue be reproduced?

_No response_

noajshu:
- don't reuse compiled Bazel outputs across CI runners.
- build Linux distribution wheels with a portable x86-64 baseline in a pinned manylinux container.
- Repair and validate Linux wheels with `auditwheel`.
- Use correct CPython ABI tags.
- Build, install, and import-test Linux x86-64 and macOS ARM64 wheels for Python 3.11-3.13 through one reusable workflow before publication.
- publish macOS arm64 only (end of support for intel macs)
- Run the same build and validation path on pull requests (of course, without uploading or publishing the wheels.)

Fixes #291
Fixes #296

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -9,6 +9,14 @@ build --cxxopt='-std=c++20'
 # native CPU tuning in //src:OPT_COPTS.
 build:wheel --define=tesseract_portable_wheel=true
 
+# macOS arm64 wheels are tagged macosx_11_0. Apply the deployment target to
+# every build action, including external dependencies and final link actions.
+build:macos_wheel --action_env=MACOSX_DEPLOYMENT_TARGET=11.0
+build:macos_wheel --host_action_env=MACOSX_DEPLOYMENT_TARGET=11.0
+build:macos_wheel --repo_env=MACOSX_DEPLOYMENT_TARGET=11.0
+build:macos_wheel --copt=-mmacosx-version-min=11.0
+build:macos_wheel --linkopt=-mmacosx-version-min=11.0
+
 # Needed to fix issues like this which arose in GCC 15 on Linux:
 # "error: relocation refers to local symbol [...], which is defined in a discarded section"
 # We keep it linux-only so macos does not receive GNU ld flags
--- a/.bazelrc
+++ b/.bazelrc
@@ -9,13 +9,21 @@ build --cxxopt='-std=c++20'
 # native CPU tuning in //src:OPT_COPTS.
 build:wheel --define=tesseract_portable_wheel=true
 
-# macOS arm64 wheels are tagged macosx_11_0. Apply the deployment target to
-# every build action, including external dependencies and final link actions.
-build:macos_wheel --action_env=MACOSX_DEPLOYMENT_TARGET=11.0
-build:macos_wheel --host_action_env=MACOSX_DEPLOYMENT_TARGET=11.0
-build:macos_wheel --repo_env=MACOSX_DEPLOYMENT_TARGET=11.0
-build:macos_wheel --copt=-mmacosx-version-min=11.0
-build:macos_wheel --linkopt=-mmacosx-version-min=11.0
+# macOS wheels enforce the deployment target matching their platform tag:
+# arm64 wheels are tagged macosx_11_0 and x86-64 wheels macosx_10_15. Apply
+# each deployment target to every build action, including external
+# dependencies and final link actions.
+build:macos_arm64_wheel --action_env=MACOSX_DEPLOYMENT_TARGET=11.0
+build:macos_arm64_wheel --host_action_env=MACOSX_DEPLOYMENT_TARGET=11.0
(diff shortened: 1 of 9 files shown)


.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,226 @@
...
+    - name: Upload Linux wheel
+      if: inputs.upload_artifacts
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      with:
+        name: python-wheels-linux-${{ matrix.python-version }}
+        path: wheelhouse/*.whl
+
+  build_macos_wheels:
+    name: macOS wheels (${{ matrix.python-version }})
+    runs-on: macos-14
```
arshpreetmaan: runs-on: macos-14 currently resolves to an ARM64 runner, so every entry in this matrix produces only an ARM64 macOS wheel. However, the wheel target still explicitly defines :macos_x86 -> macosx_10_15_x86_64 in BUILD, which indicates that Intel macOS remains a supported distribution target.

Could we make architecture part of this matrix and add an x86-64 macOS runner/build/import-validation entry for each supported Python version? Otherwise, the reusable workflow does not actually build and validate all supported macOS wheels, and no Intel wheel will be included in the prerelease or stable-release artifacts.

.bazelrc
```diff
@@ -5,6 +5,18 @@ common --enable_platform_specific_config
...
 
+# Public Linux wheels use a portable x86-64 baseline. Normal builds retain the
+# native CPU tuning in //src:OPT_COPTS.
+build:wheel --define=tesseract_portable_wheel=true
+
+# macOS arm64 wheels are tagged macosx_11_0. Apply the deployment target to
+# every build action, including external dependencies and final link actions.
+build:macos_wheel --action_env=MACOSX_DEPLOYMENT_TARGET=11.0
+build:macos_wheel --host_action_env=MACOSX_DEPLOYMENT_TARGET=11.0
+build:macos_wheel --repo_env=MACOSX_DEPLOYMENT_TARGET=11.0
```
arshpreetmaan: The deployment-target configuration also needs to be architecture-specific: the shared --config=macos_wheel currently forces macOS 11.0 globally, while the x86-64 wheel is tagged macosx_10_15_x86_64. Either the Intel wheel tag should be changed to 11.0, or separate ARM64 and x86-64 wheel configurations should enforce deployment targets consistent with their respective tags.

.github/workflows/build-wheels.yml
noajshu: Thanks, can you PTAL?

.bazelrc
noajshu: Thanks, can you PTAL?

.github/workflows/build-wheels.yml
arshpreetmaan: Thanks. With the clarified support scope that published macOS wheels are ARM64-only, this resolves my concern. The current implementation now removes the stale x86-64 wheel target instead of leaving it apparently supported but unbuilt and unvalidated.

Could you please also update the PR description, which still says that macOS ARM64/x86-64 wheels are built and tested, to say macOS ARM64 only? It would also be helpful to document that prebuilt Intel macOS wheels are not supported

.bazelrc
arshpreetmaan: Resolved

arshpreetmaan:
LGTM! Thank you @noajshu for the fixes.

arshpreetmaan (approved):

noajshu:
Thanks for the review @arshpreetmaan !!

Issue #297, opened 2026-08-10
Tesseract Sinter integrations drop low-confidence outcomes and can undercount logical failures


arshpreetmaan:
### Description

Tesseract produces a per-shot low-confidence status when decoding does not succeed confidently. The C++ CLI records these shots separately, allowing the conservative logical failure rate to be calculated as:

```
(num_errors + num_low_confidence) / num_shots
```
The Sinter integrations currently return only observable predictions and discard the low-confidence status. This affects the existing monolithic TesseractSinterCompiledDecoder and decode_via_files paths, as well as the multi-pass Sinter integration introduced in PR #256.

As a result, some low-confidence shots may be counted as successful by Sinter, causing it to underestimate the conservative logical failure rate and overestimate decoder performance.

This issue was first noticed during review of PR #256, but it predates that PR.

simbot912:
This is a really important fix for the Sinter integration! Low-confidence outcomes being silently dropped could lead to underestimating logical error rates, which would be misleading for benchmarking studies. 

The conservative approach of counting `errors + discards` as the failure count makes sense — it's better to overestimate than underestimate QEC performance. Have you considered adding a flag to let users choose between the conservative (errors + discards) and optimistic (errors only) counting modes? Some benchmarking frameworks might want both perspectives.

sdoygb:
Closing note from the Stim side: our companion PR [quantumlib/Stim#1101](https://github.com/quantumlib/Stim/pull/1101) (adding `discards_b8_out_path` to the `sinter.Decoder.decode_via_files` file protocol) has been withdrawn.

Strilanc's review made clear that extending the file-decoder protocol is not the direction Stim wants to go:
- it adds a required concept to every third-party `sinter.Decoder` implementation (breaking compatibility),
- 'record discards' is the first step on a complexity ratchet (record errors, corrections, gaps, ...), which is exactly what `sinter.Sampler` was introduced to avoid,
- file-writing itself is a performance pitfall (disk contention kills parallelism on many cores); memory is the preferred channel.

So the file-decoder protocol stays unchanged, and this issue is addressed entirely on the compiled-decoder side by #300. Thanks for the discussion.

PR #300, opened 2026-08-14
Report low-confidence Sinter decodes as discards

danielgaskins:
## Summary

- append Sinter's discard byte to compiled decoder results
- set the discard byte when Tesseract reports low confidence
- document `errors + discards` as the conservative logical failure count
- add direct-output and Sinter-sampler regression tests

This addresses the compiled-decoder portion of #297. The pinned Sinter file-decoder API reads exactly the observable bytes for each shot and has no decoder-controlled discard channel, so `decode_via_files` remains unchanged. The multipass decoder in #256 will need equivalent propagation when it is integrated.

## Testing

- `clang-format --dry-run --Werror src/tesseract_sinter_compat.pybind.h`
- focused low-confidence tests: 2 passed
- `bazel test --jobs=1 //src/...`: 12 of 13 targets passed; the pre-existing unseeded `test_sinter_decode_surface_code` exceeded its statistical threshold locally


danielgaskins:
Could someone with Actions access rerun the failed jobs? The macOS failure came from the existing unseeded `test_sinter_decode_surface_code` statistical threshold. The new deterministic discard tests passed, as did formatting and all wheel builds.

mhucka:
> Could someone with Actions access rerun the failed jobs? The macOS failure came from the existing unseeded `test_sinter_decode_surface_code` statistical threshold. The new deterministic discard tests passed, as did formatting and all wheel builds.

Done. FYI that I had to rerun it twice, actually; the first rerun saw macos succeed but linux fail; the second rerun saw the linux tests succeed. (Mentioning it in case this is useful for additional development.)

danielgaskins:
Thanks for flagging this. The compatibility tests were still using Sinter's legacy `sample_decode` helper, which treats the new discard byte as observable data. I updated them to use Sinter's discard-aware sampler. The Bazel target now passes in 10 consecutive runs.

PR #301, opened 2026-08-14
Handle strided arrays in Sinter decoding

danielgaskins:
`decode_shots_bit_packed` accepted non-contiguous NumPy views, but treated the bytes within each shot as contiguous. For a view such as `shots[:, ::2]`, this could read the wrong bytes and decode a different set of detectors.

Use pybind11's two-dimensional array accessor so both strides are respected. Add a regression test where the second byte of each shot is non-contiguous.

Tests:
- `bazel test --jobs=1 src/... //docs:tutorial_jupytext_sync_test`
- `clang-format-14 --dry-run --Werror src/tesseract_sinter_compat.pybind.h`


simbot912:
Nice catch on the strided array bug! The `shots[:, ::2]` case is a real-world scenario — many pipelines use strided views to interleave detector data from multiple sources.

A couple of thoughts on the fix:
- The pybind11 two-dimensional array accessor is the right approach. This ensures both row and column strides are respected.
- The regression test with non-contiguous shots is essential — this is exactly the kind of bug that's hard to catch without explicit test coverage.

One question: does this fix also handle Fortran-contiguous arrays (column-major), or does it specifically target C-contiguous with non-unit column strides? In some HPC pipelines, detector data comes in Fortran order from GPU kernels.

Also, is there a performance impact for the common contiguous case? The two-dimensional accessor might be slightly slower than direct byte access.

danielgaskins:
Yes, this also handles Fortran-contiguous arrays. `unchecked<2>()` uses both NumPy strides for each `(shot, byte)` access. I checked a two-byte Fortran-order input locally, and it produced the expected predictions.

For C-contiguous input, the accessor does a little more index calculation than the previous raw pointer access. I have not benchmarked the two implementations, so I do not want to claim zero overhead. The decoding work is likely to dominate, but a contiguous fast path could be added later if profiling shows this matters.

danielgaskins:
Current main now handles this in the shared `decode_sinter_shots_bit_packed` helper added by #256, with coverage for row-strided, column-strided, and Fortran-ordered inputs. That covers the original bug more broadly than this branch, so I’m closing this PR as superseded. Thanks for the reviews and branch updates.

PR #303, opened 2026-08-20
Avoid redundant DEM flattening during decoder setup

abbrazi:
Keeps the same semantics at the cost of a few deep copies, with ~1.5–2× speedup in decoder init. 
Most workflows won't notice; mine does.


noajshu (approved):
This could also be addressed upstream in Stim, but I'm happy to approve this to fix your workflow. Good luck with your JIT work!

PR #308, opened 2026-08-24
Support py14 and create a dryrun version for PRs

NoureldinYosri:
- delete the LLM created .github/workflows/build-wheels.yml
- restore (pre-) release workflows to original state
- add support for py14


LalehB (approved):
Thank you Nour!

noajshu:
Hi @NoureldinYosri . it looks like this change introduced flakiness to PR wheel CI. For example, this run fails with an Illegal instruction while executing generate_stubs in the Linux/Python 3.14 wheel build: [https://github.com/quantumlib/tesseract-decoder/actions/runs/33297979094/job/99220818139](https://github.com/quantumlib/tesseract-decoder/actions/runs/33297979094/job/99220818139)

I suspect the problem is that the new workflow builds without --config=wheel while reusing compiled Bazel outputs across hosted runners, so -march=native artifacts may not be portable between runners. Could we restore the previous build-only wheel validation workflow, or at least use --config=wheel and disable the compiled disk cache?

